PR #1001, opened 2023-06-20
refactor: Deleted redundant extension method.

zabeen:
This change is [Reviewable](https://reviewable.io/reviews/Anthony-Nolan/Atlas/1001)

### DIFF
--- a/Atlas.MatchPrediction.Data/Models/HaplotypeFrequency.cs
+++ b/Atlas.MatchPrediction.Data/Models/HaplotypeFrequency.cs
@@ -1,7 +1,5 @@
 ﻿using System.ComponentModel.DataAnnotations;
 using System.ComponentModel.DataAnnotations.Schema;
-using Atlas.Common.GeneticData;
-using Atlas.Common.GeneticData.PhenotypeInfo;
 using Atlas.Common.Public.Models.GeneticData;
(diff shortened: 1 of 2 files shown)


zabeen (commented):
> *[Reviewable](https://reviewable.io/reviews/Anthony-Nolan/Atlas/1001)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @daria-sorokina-da)

___
*[`Atlas.MatchPrediction.Data/Repositories/HaplotypeFrequenciesRepository.cs` line 115 at r1](https://reviewable.io/reviews/Anthony-Nolan/Atlas/1001#-NYNM3wO6_u0FqYwISaB:-NYNM3wO6_u0FqYwISaC:b-eujud1) ([raw file](https://github.com/Anthony-Nolan/Atlas/blob/8494defdf5fdd0e6a2376f9271eb90582dd119a3/Atlas.MatchPrediction.Data/Repositories/HaplotypeFrequenciesRepository.cs#L115)):*
> ```Smalltalk
>                     var frequencyModels = await conn.QueryAsync<HaplotypeFrequency>(sql, new {setId}, commandTimeout: 600);
>                     return frequencyModels.ToDictionary(
>                         f => f.Hla,
> ```

maybe I am missing something - I don't know why code was returning a new object with a copy of the data which is already a prop on the source object... all tests are passing so, think it may just have been an artefact?

Issue #1053, opened 2023-09-05
Dead-lettered searchable donor updates on LIVE-WMDA-ATLAS


zabeen:
There are over 500 DL updates on `matching-transient-a` sub that mostly date from 16th August 2023.

1) why did they dead-letter?
2) how do we ensure that these donors are in correct state within the active matching algorithm? The large number precludes a manual sql query.
3) how do we force updates of the affected donors so the final state is in line with the donor store?

zabeen:
### HLD

- Write two DLQ reader functions on donor import functions app - one for each `transient` sub on the `updated-searchable-donors` topic
 - Each function should batch read the DLQ ([existing example](https://github.com/Anthony-Nolan/Atlas/blob/a3a993d83cd68d064b1ffe78d02adf9dc64ae38a/Atlas.MatchPrediction.Functions/Functions/MatchPredictionFunctions.cs#L76C61-L76C61)) and extract the donor Ids
 - Ids are then passed to a service that will check against the Atlas donor store and [save updates for each ID](https://github.com/Anthony-Nolan/Atlas/blob/master/Atlas.DonorImport/Services/DonorUpdates/DonorUpdatesSaver.cs):
   - if ID exists, save a `IsAvailableForSearch: true` update 
   - if ID not found, save a `IsAvailableForSearch: false` update

zabeen:
@IgorKupreychik re: point 1 on the card, I doubt AI will have any info after so many months, so you can just focus on the dev work :)

zabeen:
### Testing Notes

1. Force `updated-searchable-donors` messages to dead-letter, by breaking the db connection string on the [donor management functions app](https://portal.azure.com/#@ukmarrow.org/resource/subscriptions/6114522f-eea5-44ab-94ab-af37ffffc4d3/resourceGroups/DEV-ATLAS-RESOURCE-GROUP/providers/Microsoft.Web/sites/DEV-ATLAS-MATCHING-DONOR-MANAGEMENT-FUNCTION/appServices).
2. Upload a donor import file with new donor and delete donor updates.
3. Once the update messages from the file import have dead-lettered, you should see the messages disappear off the DLQ and be republished to the same topic after a couple of minutes. The re-published messages should have the correct contents.
[...]

PR #1098, opened 2023-10-17
feature: #1053: Dead-lettered searchable donor updates on LIVE-WMDA-A…

IgorKupreychik:
…TLAS

### DIFF
--- a/Atlas.DonorImport/ExternalInterface/DependencyInjection/ServiceConfiguration.cs
+++ b/Atlas.DonorImport/ExternalInterface/DependencyInjection/ServiceConfiguration.cs
@@ -43,6 +43,14 @@ namespace Atlas.DonorImport.ExternalInterface.DependencyInjection
             services.RegisterImportDatabaseTypes(fetchSqlConnectionString);
         }
 
+        public static void RegisterDonorUpdateServices(
+             this IServiceCollection services,
+             Func<IServiceProvider, string> fetchDonorImportDatabaseConnectionString)
+        {
+            services.AddScoped<IDonorUpdatesSaver, DonorUpdatesSaver>();
+            services.AddScoped<IPublishableDonorUpdatesRepository>(sp => new PublishableDonorUpdatesRepository(fetchDonorImportDatabaseConnectionString(sp)));
+        }
+
         public static void RegisterDonorReader(
             this IServiceCollection services,
             Func<IServiceProvider, string> fetchDonorImportDatabaseConnectionString)
--- a/Atlas.MatchingAlgorithm.Api/Startup.cs
+++ b/Atlas.MatchingAlgorithm.Api/Startup.cs
@@ -93,7 +93,8 @@ namespace Atlas.MatchingAlgorithm.Api
                 OptionsReaderFor<NotificationsServiceBusSettings>(),
                 ConnectionStringReader("PersistentSql"),
                 ConnectionStringReader("SqlA"),
-                ConnectionStringReader("SqlB"));
+                ConnectionStringReader("SqlB"),
+                ConnectionStringReader("DonorSql"));
 
             services.ConfigureSwaggerService();
 
--- a/Atlas.MatchingAlgorithm.Api/appsettings.json
+++ b/Atlas.MatchingAlgorithm.Api/appsettings.json
@@ -63,7 +63,8 @@
   "ConnectionStrings": {
     "SqlA": "Data Source=(local);Initial Catalog=AtlasMatchingA;Integrated Security=True;MultipleActiveResultSets=True;",
     "SqlB": "Data Source=(local);Initial Catalog=AtlasMatchingB;Integrated Security=True;MultipleActiveResultSets=True;",
-    "PersistentSql": "Data Source=(local);Initial Catalog=Atlas;Integrated Security=True;MultipleActiveResultSets=True;"
+    "PersistentSql": "Data Source=(local);Initial Catalog=Atlas;Integrated Security=True;MultipleActiveResultSets=True;",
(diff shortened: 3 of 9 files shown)


zabeen (changes requested):
> Reviewed 9 of 9 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/Anthony-Nolan/Atlas/1098)* status: all files reviewed, 1 unresolved discussion (waiting on @IgorKupreychik)

___
*[`Atlas.MatchingAlgorithm.Functions.DonorManagement/Functions/DonorManagementFunctions.cs` line 82 at r1](https://reviewable.io/reviews/Anthony-Nolan/Atlas/1098#-NgxRc9g3Hhad9LJ4Ach:-NgxRc9g3Hhad9LJ4Aci:b-e5fxw3) ([raw file](https://github.com/Anthony-Nolan/Atlas/blob/e107532e44bc145b5e25ca9602919d308e1c30e8/Atlas.MatchingAlgorithm.Functions.DonorManagement/Functions/DonorManagementFunctions.cs#L82)):*
> ```Smalltalk
>             searchableDonorUpdates.ForEach(d => d.IsAvailableForSearch = existingDonors.ContainsKey(d.DonorId));
>             await donorUpdatesSaver.Save(searchableDonorUpdates);
>         }
> ```

please move this business logic to a service class so it can be unit tested

also, apologies, I was not clear on the original ticket:
when `IsAvailableForSearch` is `true` we should be publish the _latest_ info for that donor, not republish the info contained in the dead-letter. 

In the case of your code, you should create a new update object based on the info held in `existingDonors`. This is to cover the edge case of the donor data changing before DL message has been processed. E.g., if another import file was uploaded for the same donors soon after the first one.



_Code quote:_
```Smalltalk
        private async Task ProcessDeadLetterDifferentialDonorUpdates(SearchableDonorUpdate[] searchableDonorUpdates)
        {
            var existingDonors = await donorReader.GetDonors(searchableDonorUpdates.Select(d => d.DonorId));
            searchableDonorUpdates.ForEach(d => d.IsAvailableForSearch = existingDonors.ContainsKey(d.DonorId));
            await donorUpdatesSaver.Save(searchableDonorUpdates);
        }
```

zabeen (changes requested):
> Reviewed 5 of 5 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/Anthony-Nolan/Atlas/1098)* status: all files reviewed, 2 unresolved discussions (waiting on @IgorKupreychik)

___
*[`Atlas.MatchingAlgorithm/Services/DonorManagement/DonorUpdateProcessor.cs` line 152 at r2](https://reviewable.io/reviews/Anthony-Nolan/Atlas/1098#-Nh0xeEA0R7ZQAqjF1NO:-Nh0xeEA0R7ZQAqjF1NP:b54pvgf) ([raw file](https://github.com/Anthony-Nolan/Atlas/blob/5e63c61da2ffd019721a4d4bb80742a30f332196/Atlas.MatchingAlgorithm/Services/DonorManagement/DonorUpdateProcessor.cs#L152)):*
> ```Smalltalk
>         {
>             var existingDonors = await donorReader.GetDonors(searchableDonorUpdates.Select(d => d.DonorId));
>             foreach (var searchableDonorUpdate in searchableDonorUpdates)
> ```

thanks for moving the code!

my only concern is that the mutable collection being passed into the method is being modified (side-effect) rather than new objects being created from it

I don't think it's a huge deal based on the context because as of now the array being pass in is not being used anywhere else, but it is something to think about in terms of best practice
___
*[`Atlas.MatchingAlgorithm.Test/Services/DonorManagement/DonorUpdateProcessorTests.cs` line 277 at r2](https://reviewable.io/reviews/Anthony-Nolan/Atlas/1098#-Nh0yrWpBrdMjjlP-oqk:-Nh0yrWpBrdMjjlP-oql:buiznhx) ([raw file](https://github.com/Anthony-Nolan/Atlas/blob/5e63c61da2ffd019721a4d4bb80742a30f332196/Atlas.MatchingAlgorithm.Test/Services/DonorManagement/DonorUpdateProcessorTests.cs#L277)):*
> ```Smalltalk
> 
>             await donorReader.Received(1).GetDonors(Arg.Any<IEnumerable<int>>());
>             await donorUpdatesSaver.Received(1).Save(Arg.Is<IReadOnlyCollection<SearchableDonorUpdate>>(l =>
> ```

please split these asserts to their own tests, i.e., one unit test covers where donors are present in the donor store, and the other test covers case where donor is missing - that way each unit test can be clearly named after the expected outcome

IgorKupreychik:
*[`Atlas.MatchingAlgorithm/Services/DonorManagement/DonorUpdateProcessor.cs` line 152 at r2](https://reviewable.io/reviews/Anthony-Nolan/Atlas/1098#-Nh0xeEA0R7ZQAqjF1NO:-Nh1_aM4DP9Fs5toVYre:bmahhio) ([raw file](https://github.com/Anthony-Nolan/Atlas/blob/5e63c61da2ffd019721a4d4bb80742a30f332196/Atlas.MatchingAlgorithm/Services/DonorManagement/DonorUpdateProcessor.cs#L152)):*
<details><summary><i>Previously, zabeen (Dr Zabeen Patel) wrote…</i></summary><blockquote>

thanks for moving the code!

my only concern is that the mutable collection being passed into the method is being modified (side-effect) rather than new objects being created from it

I don't think it's a huge deal based on the context because as of now the array being pass in is not being used anywhere else, but it is something to think about in terms of best practice
</blockquote></details>

@zabeen, in this particular case the only thing we do is update objects in searchableDonorUpdates and save them, I don't see any reason (other than just following "best practices") to use immutable collection here, it will just increase the amount of memory used and won't give any benefits. However, if you think it should be used here either way - let me know and I'll update the code.

Issue #1271, opened 2024-04-03
Matching - Move RegistryCode filtering to earlier stage


zabeen:
Registry filtering step in `SearchService`: `var resultsFilteredByDonorDetails = donorDetailsResultFilterer.FilterResultsByDonorData` could be moved to earlier in matching

e.g., where all the other donor info filtering is done:
https://github.com/Anthony-Nolan/Atlas/blob/master/Atlas.MatchingAlgorithm/Services/Search/Matching/MatchingService.cs#L104

Or to the database layer itself, now that `RegistryCode` is being stored in the matching db `donors` table.

SergeyEzh:
@zabeen , https://github.com/Anthony-Nolan/Atlas/blob/4b91dea08eb072d585a473712864fdfea8da158b/Atlas.MatchingAlgorithm/Services/Search/Matching/DonorMatchingService.cs#L126 -- I'd propose eliminate this, `Dictionary.ContainsKey` works as fast as `HashSet.Contains` do. But taking into consideration batch size, it worth save 5Mb RAM and some CPU time spent on copying data just for fitting to type `HashSet<T>`

PR #1286, opened 2024-04-17
feature: #1271: move registry code filtering to database side

SergeyEzh:
@zabeen ,I'm creating a draft PR because I didn't manage to review tests and write new ones yet. Please review the feature itself, while I'm working on tests

- - -
This change is [Reviewable](https://reviewable.io/reviews/Anthony-Nolan/Atlas/1286)


zabeen (commented):
> Reviewed 16 of 17 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/Anthony-Nolan/Atlas/1286)* status: 16 of 17 files reviewed, 5 unresolved discussions (waiting on @SergeyEzh)

___
*[`Atlas.MatchingAlgorithm.Common/Models/Matching/MatchingFilteringOptions.cs` line 26 at r1](https://reviewable.io/reviews/Anthony-Nolan/Atlas/1286#-NvlBkgXDtjWarmCq-2J:-NvlBkgXDtjWarmCq-2K:bgwyvxc) ([raw file](https://github.com/Anthony-Nolan/Atlas/blob/fc204fdef7918ee6a4d1dcbd0133411815848f44/Atlas.MatchingAlgorithm.Common/Models/Matching/MatchingFilteringOptions.cs#L26)):*
> ```Smalltalk
>         public ICollection<int> DonorIds { get; set; }
> 
>         public bool ShouldFilterOnDonorIds => DonorIds != null;
> ```

could extend to check "not empty" condition as well



_Code quote:_
```Smalltalk
!= null
```
___
*[`Atlas.MatchingAlgorithm.Common/Models/Matching/MatchingFilteringOptions.cs` line 30 at r1](https://reviewable.io/reviews/Anthony-Nolan/Atlas/1286#-NvlBZV2BE8c_8wvq6Tb:-NvlBZV31G5DW2ptUxLx:bgwyvxc) ([raw file](https://github.com/Anthony-Nolan/Atlas/blob/fc204fdef7918ee6a4d1dcbd0133411815848f44/Atlas.MatchingAlgorithm.Common/Models/Matching/MatchingFilteringOptions.cs#L30)):*
> ```Smalltalk
>         public ICollection<string> RegistryCodes { get; set; }
> 
>         public bool ShouldFilterOnRegistryCodeIds => RegistryCodes != null;  
> ```

could extend to check "not empty" condition as well



_Code quote:_
```Smalltalk
!= null
```
___
*[`Atlas.MatchingAlgorithm.Data/Repositories/DonorRetrieval/DonorSearchRepository.cs` line 204 at r1](https://reviewable.io/reviews/Anthony-Nolan/Atlas/1286#-NvlBxdx2ncwViXQiE6x:-NvlBxdx2ncwViXQiE6y:b-jnmxso) ([raw file](https://github.com/Anthony-Nolan/Atlas/blob/fc204fdef7918ee6a4d1dcbd0133411815848f44/Atlas.MatchingAlgorithm.Data/Repositories/DonorRetrieval/DonorSearchRepository.cs#L204)):*
> ```Smalltalk
> {donorIdTempTableJoin}
> {donorUpdatedJoin}
> {donorRegistryCodeTempTableJoin}
> ```

suggest moving this down to the third sql select operation where `donorTypeFilteredJoin` is applied - that way join to `Donors` table only needs to be done once, to filter by donor type and/or registry code



_Code quote:_
```Smalltalk
{donorRegistryCodeTempTableJoin}
```
___
*[`Atlas.MatchingAlgorithm.Data/Repositories/DonorRetrieval/DonorSearchRepository.cs` line 235 at r1](https://reviewable.io/reviews/Anthony-Nolan/Atlas/1286#-NvlCgFpD6c7r5JnfwTg:-NvlCgFpD6c7r5JnfwTh:bjbb0rt) ([raw file](https://github.com/Anthony-Nolan/Atlas/blob/fc204fdef7918ee6a4d1dcbd0133411815848f44/Atlas.MatchingAlgorithm.Data/Repositories/DonorRetrieval/DonorSearchRepository.cs#L235)):*
> ```Smalltalk
>                         }
> 
>                         if (filteringOptions.DonorIds != null)
> ```

I know this isn't your code, but please replace this with `ShouldFilterOnDonorIds`



_Code quote:_
```Smalltalk
.DonorIds != null
```
___
*[`Atlas.MatchingAlgorithm.Test/TestHelpers/Builders/SearchRequests/AlleleLevelMatchCriteriaBuilder.cs` line 80 at r1](https://reviewable.io/reviews/Anthony-Nolan/Atlas/1286#-NvlE1df9gYfLzGIqcIN:-NvlE1df9gYfLzGIqcIO:b-hxajt0) ([raw file](https://github.com/Anthony-Nolan/Atlas/blob/fc204fdef7918ee6a4d1dcbd0133411815848f44/Atlas.MatchingAlgorithm.Test/TestHelpers/Builders/SearchRequests/AlleleLevelMatchCriteriaBuilder.cs#L80)):*
> ```Smalltalk
>     public class MatchCriteriaBuilder 
>     {
>         private readonly AlleleLevelMatchCriteriaBuilder inner;
> ```

if this was accessible, then wouldn't you need all the "with" methods below that aren't doing anything extra...

Or the test could first build `Allele..Builder` and then pass it into constructor of `Match..Builder` when it's ready - I don't think the criteria would need to be modified further in the scope of a single test?

It's a lot of extra code that isn't doing much :laughing: 



_Code quote:_
```Smalltalk
private readonly
```

zabeen (commented):
> Reviewed 1 of 17 files at r1.
> *[Reviewable](https://reviewable.io/reviews/Anthony-Nolan/Atlas/1286)* status: all files reviewed, 5 unresolved discussions (waiting on @SergeyEzh)

SergeyEzh (commented):
Resolved some comments. I continue work on adding tests and I'm pushing this commit let conversation about where to put `join Donor` going.

> *[Reviewable](https://reviewable.io/reviews/Anthony-Nolan/Atlas/1286#-:-Nw4dcMJ9zLOp2LClwlZ:b7wgyg0)* status: 11 of 17 files reviewed, 6 unresolved discussions (waiting on @zabeen)

___
*[`Atlas.MatchingAlgorithm.Common/Models/Matching/MatchingFilteringOptions.cs` line 26 at r1](https://reviewable.io/reviews/Anthony-Nolan/Atlas/1286#-NvlBkgXDtjWarmCq-2J:-Nw4S0L3CWKgVWAIk7aa:b-896fix) ([raw file](https://github.com/Anthony-Nolan/Atlas/blob/fc204fdef7918ee6a4d1dcbd0133411815848f44/Atlas.MatchingAlgorithm.Common/Models/Matching/MatchingFilteringOptions.cs#L26)):*
<details><summary><i>Previously, zabeen (Dr Zabeen Patel) wrote…</i></summary><blockquote>

could extend to check "not empty" condition as well
</blockquote></details>

Done.
___
*[`Atlas.MatchingAlgorithm.Common/Models/Matching/MatchingFilteringOptions.cs` line 30 at r1](https://reviewable.io/reviews/Anthony-Nolan/Atlas/1286#-NvlBZV2BE8c_8wvq6Tb:-Nw4S0ac4pulzZUDq5fA:b-896fix) ([raw file](https://github.com/Anthony-Nolan/Atlas/blob/fc204fdef7918ee6a4d1dcbd0133411815848f44/Atlas.MatchingAlgorithm.Common/Models/Matching/MatchingFilteringOptions.cs#L30)):*
<details><summary><i>Previously, zabeen (Dr Zabeen Patel) wrote…</i></summary><blockquote>

could extend to check "not empty" condition as well
</blockquote></details>

Done.
___
*[`Atlas.MatchingAlgorithm.Data/Repositories/DonorRetrieval/DonorSearchRepository.cs` line 204 at r1](https://reviewable.io/reviews/Anthony-Nolan/Atlas/1286#-NvlBxdx2ncwViXQiE6x:-Nw4Mxdq1tEyD9VVHj4M:b-5iv6ka) ([raw file](https://github.com/Anthony-Nolan/Atlas/blob/fc204fdef7918ee6a4d1dcbd0133411815848f44/Atlas.MatchingAlgorithm.Data/Repositories/DonorRetrieval/DonorSearchRepository.cs#L204)):*
<details><summary><i>Previously, zabeen (Dr Zabeen Patel) wrote…</i></summary><blockquote>

suggest moving this down to the third sql select operation where `donorTypeFilteredJoin` is applied - that way join to `Donors` table only needs to be done once, to filter by donor type and/or registry code
</blockquote></details>

I'd rather move donor type filtering to first & second queries because it isn't simple queries but inserting in temp tables. Filtering data is cheaper than inserting, so if we move that in upper queries, then we will be moving less data back and forth inside sql server.
___
*[`Atlas.MatchingAlgorithm.Data/Repositories/DonorRetrieval/DonorSearchRepository.cs` line 217 at r1](https://reviewable.io/reviews/Anthony-Nolan/Atlas/1286#-NvlFQuX-eYkN9ONGpBg:-Nw4dUJd11bfQHmRDhen:b2ss7xm) ([raw file](https://github.com/Anthony-Nolan/Atlas/blob/fc204fdef7918ee6a4d1dcbd0133411815848f44/Atlas.MatchingAlgorithm.Data/Repositories/DonorRetrieval/DonorSearchRepository.cs#L217)):*
<details><summary><i>Previously, zabeen (Dr Zabeen Patel) wrote…</i></summary><blockquote>

as above
</blockquote></details>

see comment above
___
*[`Atlas.MatchingAlgorithm.Data/Repositories/DonorRetrieval/DonorSearchRepository.cs` line 235 at r1](https://reviewable.io/reviews/Anthony-Nolan/Atlas/1286#-NvlCgFpD6c7r5JnfwTg:-Nw4dRMYDk1i45PhWEXV:b-896fix) ([raw file](https://github.com/Anthony-Nolan/Atlas/blob/fc204fdef7918ee6a4d1dcbd0133411815848f44/Atlas.MatchingAlgorithm.Data/Repositories/DonorRetrieval/DonorSearchRepository.cs#L235)):*
<details><summary><i>Previously, zabeen (Dr Zabeen Patel) wrote…</i></summary><blockquote>

I know this isn't your code, but please replace this with `ShouldFilterOnDonorIds`
</blockquote></details>

Done.
___
*[`Atlas.MatchingAlgorithm.Test/TestHelpers/Builders/SearchRequests/AlleleLevelMatchCriteriaBuilder.cs` line 80 at r1](https://reviewable.io/reviews/Anthony-Nolan/Atlas/1286#-NvlE1df9gYfLzGIqcIN:-Nw4dX6rCHXkL4fD1vdI:bnctwun) ([raw file](https://github.com/Anthony-Nolan/Atlas/blob/fc204fdef7918ee6a4d1dcbd0133411815848f44/Atlas.MatchingAlgorithm.Test/TestHelpers/Builders/SearchRequests/AlleleLevelMatchCriteriaBuilder.cs#L80)):*
<details><summary><i>Previously, zabeen (Dr Zabeen Patel) wrote…</i></summary><blockquote>

if this was accessible, then wouldn't you need all the "with" methods below that aren't doing anything extra...

Or the test could first build `Allele..Builder` and then pass it into constructor of `Match..Builder` when it's ready - I don't think the criteria would need to be modified further in the scope of a single test?

It's a lot of extra code that isn't doing much :laughing:
</blockquote></details>

I wanted to avoid massive tests update.

zabeen (commented):
> Reviewed 4 of 7 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/Anthony-Nolan/Atlas/1286)* status: 15 of 18 files reviewed, 2 unresolved discussions (waiting on @daria-sorokina-da)

___
*[`Atlas.MatchingAlgorithm.Data/Repositories/DonorRetrieval/DonorSearchRepository.cs` line 204 at r1](https://reviewable.io/reviews/Anthony-Nolan/Atlas/1286#-NvlBxdx2ncwViXQiE6x:-NwFacpj7Trt4xtXsrqx:b18helw) ([raw file](https://github.com/Anthony-Nolan/Atlas/blob/fc204fdef7918ee6a4d1dcbd0133411815848f44/Atlas.MatchingAlgorithm.Data/Repositories/DonorRetrieval/DonorSearchRepository.cs#L204)):*
<details><summary><i>Previously, SergeyEzh wrote…</i></summary><blockquote>

I'd rather move donor type filtering to first & second queries because it isn't simple queries but inserting in temp tables. Filtering data is cheaper than inserting, so if we move that in upper queries, then we will be moving less data back and forth inside sql server.
</blockquote></details>

@daria-sorokina-da do you have any thoughts?

daria-sorokina-da (commented):
> Reviewed 11 of 17 files at r1, 7 of 7 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/Anthony-Nolan/Atlas/1286)* status: all files reviewed, 4 unresolved discussions (waiting on @SergeyEzh and @zabeen)

___
*[`Atlas.MatchingAlgorithm.Common/Models/Matching/MatchingFilteringOptions.cs` line 24 at r2](https://reviewable.io/reviews/Anthony-Nolan/Atlas/1286#-NwKk8K8B_mEOjwJPfA4:-NwKk8K8B_mEOjwJPfA5:b-qmyu92) ([raw file](https://github.com/Anthony-Nolan/Atlas/blob/521a96395a86daad4658e212ab1007c1aaf4f013/Atlas.MatchingAlgorithm.Common/Models/Matching/MatchingFilteringOptions.cs#L24)):*
> ```Smalltalk
>         /// Otherwise, allows all donors. 
>         /// </summary>
>         public ICollection<int> DonorIds { get; set; }
> ```

This property is nullable, right? 
Do we not mark it as nullable here because nullable are not allowed in this project?
___
*[`Atlas.MatchingAlgorithm.Common/Models/Matching/MatchingFilteringOptions.cs` line 24 at r2](https://reviewable.io/reviews/Anthony-Nolan/Atlas/1286#-NwKkThBAtqocCHU0-OH:-NwKkThBAtqocCHU0-OI:b-ezjzde) ([raw file](https://github.com/Anthony-Nolan/Atlas/blob/4b91dea08eb072d585a473712864fdfea8da158b/Atlas.MatchingAlgorithm.Common/Models/Matching/MatchingFilteringOptions.cs#L24)):*
> ```Smalltalk
>         /// Otherwise, allows all donors. 
>         /// </summary>
>         public HashSet<int> DonorIds { get; set; }
> ```

Why are we changing the type from HasSet to ICollection?
___
*[`Atlas.MatchingAlgorithm.Common/Models/Matching/MatchingFilteringOptions.cs` line 28 at r2](https://reviewable.io/reviews/Anthony-Nolan/Atlas/1286#-NwKjxhLC1WBq2ZXthcj:-NwKjxhLC1WBq2ZXthck:b-qmyu92) ([raw file](https://github.com/Anthony-Nolan/Atlas/blob/521a96395a86daad4658e212ab1007c1aaf4f013/Atlas.MatchingAlgorithm.Common/Models/Matching/MatchingFilteringOptions.cs#L28)):*
> ```Smalltalk
>         public bool ShouldFilterOnDonorIds => DonorIds != null && DonorIds.Count > 0;
> 
>         public ICollection<string> RegistryCodes { get; set; }
> ```

This property is nullable, right? 
Do we not mark it as nullable here because nullable are not allowed in this project?
___
*[`Atlas.MatchingAlgorithm.Data/Repositories/DonorRetrieval/DonorSearchRepository.cs` line 204 at r1](https://reviewable.io/reviews/Anthony-Nolan/Atlas/1286#-NvlBxdx2ncwViXQiE6x:-NwKlERk4S8uMGRc3Yz5:b-yoxkhp) ([raw file](https://github.com/Anthony-Nolan/Atlas/blob/fc204fdef7918ee6a4d1dcbd0133411815848f44/Atlas.MatchingAlgorithm.Data/Repositories/DonorRetrieval/DonorSearchRepository.cs#L204)):*
<details><summary><i>Previously, zabeen (Dr Zabeen Patel) wrote…</i></summary><blockquote>

@daria-sorokina-da do you have any thoughts?
</blockquote></details>

I think we need to be consistent, and have filtering by donor type and donor registry code together.

Whether we're doing it in the selects 1/2 or the third would depend on whether we have corresponding indexes.

Is there an index in Donor d that covers Registry Code and Donor Type?

SergeyEzh (commented):
> *[Reviewable](https://reviewable.io/reviews/Anthony-Nolan/Atlas/1286)* status: 15 of 18 files reviewed, 4 unresolved discussions (waiting on @daria-sorokina-da and @zabeen)

___
*[`Atlas.MatchingAlgorithm.Common/Models/Matching/MatchingFilteringOptions.cs` line 24 at r2](https://reviewable.io/reviews/Anthony-Nolan/Atlas/1286#-NwKk8K8B_mEOjwJPfA4:-O09DJy7Bs1vX43gcRIx:b-i1icgw) ([raw file](https://github.com/Anthony-Nolan/Atlas/blob/521a96395a86daad4658e212ab1007c1aaf4f013/Atlas.MatchingAlgorithm.Common/Models/Matching/MatchingFilteringOptions.cs#L24)):*

This property is nullable, right? 
Do we not mark it as nullable here because nullable are not allowed in this project?
</blockquote></details>

marked as nullable
___
*[`Atlas.MatchingAlgorithm.Common/Models/Matching/MatchingFilteringOptions.cs` line 24 at r2](https://reviewable.io/reviews/Anthony-Nolan/Atlas/1286#-NwKkThBAtqocCHU0-OH:-O09D_fM5xOnm3zcQh6V:b-b9lu6x) ([raw file](https://github.com/Anthony-Nolan/Atlas/blob/4b91dea08eb072d585a473712864fdfea8da158b/Atlas.MatchingAlgorithm.Common/Models/Matching/MatchingFilteringOptions.cs#L24)):*

Why are we changing the type from HasSet to ICollection?
</blockquote></details>

Its HashSet features isn't used, so I changed to more general type. Also I changed code creating this class to create List<T> instead of HashSet which is more lightweight
___
*[`Atlas.MatchingAlgorithm.Common/Models/Matching/MatchingFilteringOptions.cs` line 28 at r2](https://reviewable.io/reviews/Anthony-Nolan/Atlas/1286#-NwKjxhLC1WBq2ZXthcj:-O09E8Me-pgweSylIePw:b-i1icgw) ([raw file](https://github.com/Anthony-Nolan/Atlas/blob/521a96395a86daad4658e212ab1007c1aaf4f013/Atlas.MatchingAlgorithm.Common/Models/Matching/MatchingFilteringOptions.cs#L28)):*

This property is nullable, right? 
Do we not mark it as nullable here because nullable are not allowed in this project?
</blockquote></details>

marked as nullable
___
*[`Atlas.MatchingAlgorithm.Data/Repositories/DonorRetrieval/DonorSearchRepository.cs` line 204 at r1](https://reviewable.io/reviews/Anthony-Nolan/Atlas/1286#-NvlBxdx2ncwViXQiE6x:-O09EDby6Th7HraFN3Zp:b-n9jr9d) ([raw file](https://github.com/Anthony-Nolan/Atlas/blob/fc204fdef7918ee6a4d1dcbd0133411815848f44/Atlas.MatchingAlgorithm.Data/Repositories/DonorRetrieval/DonorSearchRepository.cs#L204)):*

I think we need to be consistent, and have filtering by donor type and donor registry code together.

Whether we're doing it in the selects 1/2 or the third would depend on whether we have corresponding indexes.

Is there an index in Donor d that covers Registry Code and Donor Type?
</blockquote></details>

I propose to move DonorType filtering to first query and add index on RegistryCode and DonorType (indexes for each RegistryCode and DonotType already exists but not both). Moving filtering to first query will decrease amount of data inserted into temp tables and, therefore decrease IO count.

daria-sorokina-da (commented):
> Reviewed 3 of 3 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/Anthony-Nolan/Atlas/1286)* status: all files reviewed, 6 unresolved discussions (waiting on @SergeyEzh and @zabeen)

*[a discussion](https://reviewable.io/reviews/Anthony-Nolan/Atlas/1286#-O09Gl7yBteFzoOXknh7:-O09Gl7yBteFzoOXknh8:baspq92) (no related file):*
Please target this PR to rc/2.3.0

___
*[a discussion](https://reviewable.io/reviews/Anthony-Nolan/Atlas/1286#-O09GpFC86jNM717EpIb:-O09GpFC86jNM717EpIc:bzgxb50) (no related file):*
Please update the versions to 2.3.0

___
___
*[`Atlas.MatchingAlgorithm.Common/Models/Matching/MatchingFilteringOptions.cs` line 26 at r3](https://reviewable.io/reviews/Anthony-Nolan/Atlas/1286#-O09IyQv2He9dRsEJ4o5:-O09IyQv2He9dRsEJ4o6:bcnsdea) ([raw file](https://github.com/Anthony-Nolan/Atlas/blob/c71cedf92448233dbd5c8c1ec224cf381fa9917a/Atlas.MatchingAlgorithm.Common/Models/Matching/MatchingFilteringOptions.cs#L26)):*
> ```Smalltalk
>         public ICollection<int>? DonorIds { get; set; }
> 
>         public bool ShouldFilterOnDonorIds => DonorIds != null && DonorIds.Count > 0;
> ```

DonorIds.Count == 0 - I would expect the filtering to work and return 0 results, why do we skip filtering in this case?
___
*[`Atlas.MatchingAlgorithm.Common/Models/Matching/MatchingFilteringOptions.cs` line 30 at r3](https://reviewable.io/reviews/Anthony-Nolan/Atlas/1286#-O09J7OAAOj1ZVDnsGLV:-O09J7OAAOj1ZVDnsGLW:bvavgmd) ([raw file](https://github.com/Anthony-Nolan/Atlas/blob/c71cedf92448233dbd5c8c1ec224cf381fa9917a/Atlas.MatchingAlgorithm.Common/Models/Matching/MatchingFilteringOptions.cs#L30)):*
> ```Smalltalk
>         public ICollection<string>? RegistryCodes { get; set; }
> 
>         public bool ShouldFilterOnRegistryCodeIds => RegistryCodes != null && RegistryCodes.Count > 0;
> ```

RegistryCodes.Count == 0 - I would expect the filtering to work and return 0 results, why do we skip filtering in this case?
___
*[`Atlas.MatchingAlgorithm.Data/Repositories/DonorRetrieval/DonorSearchRepository.cs` line 190 at r3](https://reviewable.io/reviews/Anthony-Nolan/Atlas/1286#-O09KJK-AsFX-5bd0N2-:-O09JHTeBXC_8opElD3g:bkzwekl) ([raw file](https://github.com/Anthony-Nolan/Atlas/blob/c71cedf92448233dbd5c8c1ec224cf381fa9917a/Atlas.MatchingAlgorithm.Data/Repositories/DonorRetrieval/DonorSearchRepository.cs#L190)):*
> ```Smalltalk
> 
>                     var donorRegistryCodeTempTableJoin = filteringOptions.ShouldFilterOnRegistryCodeIds 
>                         ? $" JOIN Donors d on m.DonorId = d.DonorId {donorRegistryCodeTempTableJoinConfig.FilteredJoinQueryString}" 
> ```

Instead of a separate join with Donors d, would it be better to add this to the same JOIN where we have filtering by DonorType? Won't they be working together?

Then we can add an index - (DonorType, RegistryCode) with included DonorId - and we'll start using it instead of (DonorType) with included DonorId?

daria-sorokina-da (commented):
> *[Reviewable](https://reviewable.io/reviews/Anthony-Nolan/Atlas/1286)* status: all files reviewed, 6 unresolved discussions (waiting on @SergeyEzh and @zabeen)

___
*[`Atlas.MatchingAlgorithm.Data/Repositories/DonorRetrieval/DonorSearchRepository.cs` line 204 at r1](https://reviewable.io/reviews/Anthony-Nolan/Atlas/1286#-NvlBxdx2ncwViXQiE6x:-O09LnkU7KpEZJ4uG4W7:brt23ad) ([raw file](https://github.com/Anthony-Nolan/Atlas/blob/fc204fdef7918ee6a4d1dcbd0133411815848f44/Atlas.MatchingAlgorithm.Data/Repositories/DonorRetrieval/DonorSearchRepository.cs#L204)):*
<details><summary><i>Previously, SergeyEzh wrote…</i></summary><blockquote>

I propose to move DonorType filtering to first query and add index on RegistryCode and DonorType (indexes for each RegistryCode and DonotType already exists but not both). Moving filtering to first query will decrease amount of data inserted into temp tables and, therefore decrease IO count.
</blockquote></details>

I don't see an index by RegistruCode, only by ExternalDonorCode - where do you see it?

SergeyEzh (commented):
> *[Reviewable](https://reviewable.io/reviews/Anthony-Nolan/Atlas/1286)* status: all files reviewed, 6 unresolved discussions (waiting on @daria-sorokina-da and @zabeen)

___
*[`Atlas.MatchingAlgorithm.Common/Models/Matching/MatchingFilteringOptions.cs` line 26 at r3](https://reviewable.io/reviews/Anthony-Nolan/Atlas/1286#-O09IyQv2He9dRsEJ4o5:-O09ZOX758bIjGOAqXmS:b-5spbca) ([raw file](https://github.com/Anthony-Nolan/Atlas/blob/c71cedf92448233dbd5c8c1ec224cf381fa9917a/Atlas.MatchingAlgorithm.Common/Models/Matching/MatchingFilteringOptions.cs#L26)):*

DonorIds.Count == 0 - I would expect the filtering to work and return 0 results, why do we skip filtering in this case?
</blockquote></details>

It was existing code. I didn't want to change this convention.
___
*[`Atlas.MatchingAlgorithm.Common/Models/Matching/MatchingFilteringOptions.cs` line 30 at r3](https://reviewable.io/reviews/Anthony-Nolan/Atlas/1286#-O09J7OAAOj1ZVDnsGLV:-O09Zogr6I2LN3YE3amO:b2jc46) ([raw file](https://github.com/Anthony-Nolan/Atlas/blob/c71cedf92448233dbd5c8c1ec224cf381fa9917a/Atlas.MatchingAlgorithm.Common/Models/Matching/MatchingFilteringOptions.cs#L30)):*

RegistryCodes.Count == 0 - I would expect the filtering to work and return 0 results, why do we skip filtering in this case?
</blockquote></details>

Decided to stick with existing convention here.
___
*[`Atlas.MatchingAlgorithm.Data/Repositories/DonorRetrieval/DonorSearchRepository.cs` line 204 at r1](https://reviewable.io/reviews/Anthony-Nolan/Atlas/1286#-NvlBxdx2ncwViXQiE6x:-O09_UmM5hgxShU0h8U_:b-1ouv8p) ([raw file](https://github.com/Anthony-Nolan/Atlas/blob/fc204fdef7918ee6a4d1dcbd0133411815848f44/Atlas.MatchingAlgorithm.Data/Repositories/DonorRetrieval/DonorSearchRepository.cs#L204)):*

I don't see an index by RegistruCode, only by ExternalDonorCode - where do you see it?
</blockquote></details>

oh, my bad. Confused it with ExternalDonorCode's index
___
*[`Atlas.MatchingAlgorithm.Data/Repositories/DonorRetrieval/DonorSearchRepository.cs` line 190 at r3](https://reviewable.io/reviews/Anthony-Nolan/Atlas/1286#-O09KJK-AsFX-5bd0N2-:-O09Zz8I7Pr32IOCDwSY:b-lbr5yr) ([raw file](https://github.com/Anthony-Nolan/Atlas/blob/c71cedf92448233dbd5c8c1ec224cf381fa9917a/Atlas.MatchingAlgorithm.Data/Repositories/DonorRetrieval/DonorSearchRepository.cs#L190)):*

Instead of a separate join with Donors d, would it be better to add this to the same JOIN where we have filtering by DonorType? Won't they be working together?

Then we can add an index - (DonorType, RegistryCode) with included DonorId - and we'll start using it instead of (DonorType) with included DonorId?
</blockquote></details>

I dig more into this: this method currently is never called with both DonorType and RegistryCode filters applied. As far as I understand, there's no need to do that: once we've got donor ids to filter out donors we don't need to filter by registry code.

see. https://github.com/Anthony-Nolan/Atlas/blob/7e150a45e603eaa061fa98b8b9cd4e65c9e04b65/Atlas.MatchingAlgorithm/Services/Search/Matching/DonorMatchingService.cs#L92

daria-sorokina-da (commented):
> *[Reviewable](https://reviewable.io/reviews/Anthony-Nolan/Atlas/1286)* status: all files reviewed, 6 unresolved discussions (waiting on @SergeyEzh and @zabeen)

___
*[`Atlas.MatchingAlgorithm.Common/Models/Matching/MatchingFilteringOptions.cs` line 26 at r3](https://reviewable.io/reviews/Anthony-Nolan/Atlas/1286#-O09IyQv2He9dRsEJ4o5:-O0ABA5P4OW9xxfmT0ak:b64hx4d) ([raw file](https://github.com/Anthony-Nolan/Atlas/blob/c71cedf92448233dbd5c8c1ec224cf381fa9917a/Atlas.MatchingAlgorithm.Common/Models/Matching/MatchingFilteringOptions.cs#L26)):*
<details><summary><i>Previously, SergeyEzh wrote…</i></summary><blockquote>

It was existing code. I didn't want to change this convention.
</blockquote></details>

But this particular DonorIds.Count > 0 check was added by you, right? See in the stable version:
https://github.com/Anthony-Nolan/Atlas/blob/stable/2.1.0/Atlas.MatchingAlgorithm.Common/Models/Matching/MatchingFilteringOptions.cs#L26
___
*[`Atlas.MatchingAlgorithm.Common/Models/Matching/MatchingFilteringOptions.cs` line 30 at r3](https://reviewable.io/reviews/Anthony-Nolan/Atlas/1286#-O09J7OAAOj1ZVDnsGLV:-O0ABb9K0oXm8TgC6ixz:b-d7bpdc) ([raw file](https://github.com/Anthony-Nolan/Atlas/blob/c71cedf92448233dbd5c8c1ec224cf381fa9917a/Atlas.MatchingAlgorithm.Common/Models/Matching/MatchingFilteringOptions.cs#L30)):*
<details><summary><i>Previously, SergeyEzh wrote…</i></summary><blockquote>

Decided to stick with existing convention here.
</blockquote></details>

Same as above - the existing convention is:
https://github.com/Anthony-Nolan/Atlas/blob/stable/2.1.0/Atlas.MatchingAlgorithm.Common/Models/Matching/MatchingFilteringOptions.cs#L26

And the same for DonorType (which you left untouched: https://github.com/Anthony-Nolan/Atlas/blob/stable/2.1.0/Atlas.MatchingAlgorithm.Common/Models/Matching/MatchingFilteringOptions.cs#L18
___
*[`Atlas.MatchingAlgorithm.Data/Repositories/DonorRetrieval/DonorSearchRepository.cs` line 190 at r3](https://reviewable.io/reviews/Anthony-Nolan/Atlas/1286#-O09KJK-AsFX-5bd0N2-:-O0ABmy94H7n9R6xqOlG:b6tw3a9) ([raw file](https://github.com/Anthony-Nolan/Atlas/blob/c71cedf92448233dbd5c8c1ec224cf381fa9917a/Atlas.MatchingAlgorithm.Data/Repositories/DonorRetrieval/DonorSearchRepository.cs#L190)):*
<details><summary><i>Previously, SergeyEzh wrote…</i></summary><blockquote>

I dig more into this: this method currently is never called with both DonorType and RegistryCode filters applied. As far as I understand, there's no need to do that: once we've got donor ids to filter out donors we don't need to filter by registry code.

see. https://github.com/Anthony-Nolan/Atlas/blob/7e150a45e603eaa061fa98b8b9cd4e65c9e04b65/Atlas.MatchingAlgorithm/Services/Search/Matching/DonorMatchingService.cs#L92
</blockquote></details>

At which point do we filter by DonorType? I'd expect this to happen before we know DonorIds, right? Otherwise lots of unneeded donors will be returned.

I'd expect the filtering by RegistryCode to happen at the same time.

Let's discuss.

SergeyEzh (commented):
> *[Reviewable](https://reviewable.io/reviews/Anthony-Nolan/Atlas/1286)* status: 14 of 29 files reviewed, 6 unresolved discussions (waiting on @daria-sorokina-da and @zabeen)

*[a discussion](https://reviewable.io/reviews/Anthony-Nolan/Atlas/1286#-O09Gl7yBteFzoOXknh7:-O0wxHOO8BIFfs4QqSWh:bb7f2mf) (no related file):*

Please target this PR to rc/2.3.0
</blockquote></details>

done

___
*[a discussion](https://reviewable.io/reviews/Anthony-Nolan/Atlas/1286#-O09GpFC86jNM717EpIb:-O0wxHxc5fOYwxc6Avle:bb7f2mf) (no related file):*

Please update the versions to 2.3.0
</blockquote></details>

done

___
___
*[`Atlas.MatchingAlgorithm.Common/Models/Matching/MatchingFilteringOptions.cs` line 26 at r3](https://reviewable.io/reviews/Anthony-Nolan/Atlas/1286#-O09IyQv2He9dRsEJ4o5:-O0DY4IiEbmSbSpr8mEA:b-yx7ka0) ([raw file](https://github.com/Anthony-Nolan/Atlas/blob/c71cedf92448233dbd5c8c1ec224cf381fa9917a/Atlas.MatchingAlgorithm.Common/Models/Matching/MatchingFilteringOptions.cs#L26)):*

But this particular DonorIds.Count > 0 check was added by you, right? See in the stable version:
https://github.com/Anthony-Nolan/Atlas/blob/stable/2.1.0/Atlas.MatchingAlgorithm.Common/Models/Matching/MatchingFilteringOptions.cs#L26
</blockquote></details>

removed as discussed.
___
*[`Atlas.MatchingAlgorithm.Common/Models/Matching/MatchingFilteringOptions.cs` line 30 at r3](https://reviewable.io/reviews/Anthony-Nolan/Atlas/1286#-O09J7OAAOj1ZVDnsGLV:-O0DYVDXAbV9S_au91qk:b-yx7ka0) ([raw file](https://github.com/Anthony-Nolan/Atlas/blob/c71cedf92448233dbd5c8c1ec224cf381fa9917a/Atlas.MatchingAlgorithm.Common/Models/Matching/MatchingFilteringOptions.cs#L30)):*

Same as above - the existing convention is:
https://github.com/Anthony-Nolan/Atlas/blob/stable/2.1.0/Atlas.MatchingAlgorithm.Common/Models/Matching/MatchingFilteringOptions.cs#L26

And the same for DonorType (which you left untouched: https://github.com/Anthony-Nolan/Atlas/blob/stable/2.1.0/Atlas.MatchingAlgorithm.Common/Models/Matching/MatchingFilteringOptions.cs#L18
</blockquote></details>

removed as discussed.
___
*[`Atlas.MatchingAlgorithm.Data/Repositories/DonorRetrieval/DonorSearchRepository.cs` line 190 at r3](https://reviewable.io/reviews/Anthony-Nolan/Atlas/1286#-O09KJK-AsFX-5bd0N2-:-O0ww1Nf1ILnssn6V7yW:bh7ti5k) ([raw file](https://github.com/Anthony-Nolan/Atlas/blob/c71cedf92448233dbd5c8c1ec224cf381fa9917a/Atlas.MatchingAlgorithm.Data/Repositories/DonorRetrieval/DonorSearchRepository.cs#L190)):*

At which point do we filter by DonorType? I'd expect this to happen before we know DonorIds, right? Otherwise lots of unneeded donors will be returned.

I'd expect the filtering by RegistryCode to happen at the same time.

Let's discuss.
</blockquote></details>

Reworked as it was discussed:
- Always filtering by `DonorType`
- Doing DonorType and RegistryCopy filtering in first two queries (i.e. filling temp tables stage) as it appears to be 2x faster

daria-sorokina-da (commented):
> Reviewed 15 of 15 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/Anthony-Nolan/Atlas/1286)* status: all files reviewed, 8 unresolved discussions (waiting on @SergeyEzh and @zabeen)

___
*[`Atlas.MatchingAlgorithm/Services/Search/Matching/PerLocusDonorMatchingService.cs` line 69 at r4](https://reviewable.io/reviews/Anthony-Nolan/Atlas/1286#-O12A7qx3P9dkbTObvYi:-O12A7qx3P9dkbTObvYj:bf9z9p3) ([raw file](https://github.com/Anthony-Nolan/Atlas/blob/921e7451bd3ee72b7d2ce5db19a2d3e441c96abf/Atlas.MatchingAlgorithm/Services/Search/Matching/PerLocusDonorMatchingService.cs#L69)):*
> ```Smalltalk
>             var filteringOptions = new MatchingFilteringOptions
>             {
>                 DonorType = searchType, // databaseFilteringAnalyser.ShouldFilterOnDonorTypeInDatabase(repoCriteria) ? searchType : (DonorType?) null,
> ```

do we need the comment?
___
*[`Atlas.MatchingAlgorithm.Data/Migrations/20240703115406_AddDonorTypeIndexToDonorsTable.cs` line 12 at r4](https://reviewable.io/reviews/Anthony-Nolan/Atlas/1286#-O12APup2hCjqzVPeFvB:-O12APup2hCjqzVPeFvC:brar2ne) ([raw file](https://github.com/Anthony-Nolan/Atlas/blob/921e7451bd3ee72b7d2ce5db19a2d3e441c96abf/Atlas.MatchingAlgorithm.Data/Migrations/20240703115406_AddDonorTypeIndexToDonorsTable.cs#L12)):*
> ```Smalltalk
>         protected override void Up(MigrationBuilder migrationBuilder)
>         {
>             migrationBuilder.CreateIndex(
> ```

Let's change this migration to first check whether the index is not there yet, and only if it's not yet there - add it.
I think we'll have to create it ahead of time as it may take a while to create
___
*[`Atlas.MatchingAlgorithm.Data/Migrations/20240703115406_AddDonorTypeIndexToDonorsTable.cs` line 13 at r4](https://reviewable.io/reviews/Anthony-Nolan/Atlas/1286#-O12Aafp4GtmmN3Lg1cg:-O12Aafp4GtmmN3Lg1ch:bduj1f4) ([raw file](https://github.com/Anthony-Nolan/Atlas/blob/921e7451bd3ee72b7d2ce5db19a2d3e441c96abf/Atlas.MatchingAlgorithm.Data/Migrations/20240703115406_AddDonorTypeIndexToDonorsTable.cs#L13)):*
> ```Smalltalk
>         {
>             migrationBuilder.CreateIndex(
>                 name: "IX_Donors_DonorType",
> ```

The name should IX_Donors_DonorType_RegistryCode, I think
___
*[`Atlas.MatchingAlgorithm.Data/Migrations/20240703115406_AddDonorTypeIndexToDonorsTable.cs` line 15 at r4](https://reviewable.io/reviews/Anthony-Nolan/Atlas/1286#-O12AqVEElvrSUHUWGUW:-O12AqVEElvrSUHUWGUX:b1pmo2h) ([raw file](https://github.com/Anthony-Nolan/Atlas/blob/921e7451bd3ee72b7d2ce5db19a2d3e441c96abf/Atlas.MatchingAlgorithm.Data/Migrations/20240703115406_AddDonorTypeIndexToDonorsTable.cs#L15)):*
> ```Smalltalk
>                 name: "IX_Donors_DonorType",
>                 table: "Donors",
>                 column: "DonorType")
> ```

Shouldn't the index be on two columns - DonorType and RegistryCode?
Making an index only by DonorType is not very helpful, because there are just two values in DonorType - so it's like indexing on a boolean value. It would speed up queries for Cords, i.e. if there are much less cords than adult donors.
DonorType will be less selective than RegistryCode, but it will always be provided in the query.

I'd suggest to have an index on (DonorType, RegistryCode) include (DonorId) - then we'll cover both scenarios
___
*[`Atlas.MatchingAlgorithm.Data/Migrations/20240703115406_AddDonorTypeIndexToDonorsTable.cs` line 16 at r4](https://reviewable.io/reviews/Anthony-Nolan/Atlas/1286#-O12BKqf-PnyhSachwBa:-O12BKqf-PnyhSachwBb:btf2hua) ([raw file](https://github.com/Anthony-Nolan/Atlas/blob/921e7451bd3ee72b7d2ce5db19a2d3e441c96abf/Atlas.MatchingAlgorithm.Data/Migrations/20240703115406_AddDonorTypeIndexToDonorsTable.cs#L16)):*
> ```Smalltalk
>                 table: "Donors",
>                 column: "DonorType")
>                 .Annotation("SqlServer:Include", new[] { "DonorId", "RegistryCode" });
> ```

We're only returning DonorId by this filter, right? That's why we don't need to include other columns?
___
*[`Atlas.MatchingAlgorithm.Data/Repositories/DonorRetrieval/DonorSearchRepository.cs` line 146 at r4](https://reviewable.io/reviews/Anthony-Nolan/Atlas/1286#-O12C9DZ5yakWMRJacv5:-O12C9DZ5yakWMRJacv6:b-3dl0w3) ([raw file](https://github.com/Anthony-Nolan/Atlas/blob/921e7451bd3ee72b7d2ce5db19a2d3e441c96abf/Atlas.MatchingAlgorithm.Data/Repositories/DonorRetrieval/DonorSearchRepository.cs#L146)):*
> ```Smalltalk
>                 {
>                     // Ensure we have a Donor ID even when there is only one match at this locus. 
>                     const string selectDonorIdStatement = @"CASE WHEN DonorId1 IS NULL THEN DonorId2 ELSE DonorId1 END";
> ```

Do we still need selectDonorIdStatement variable?
___
*[`Atlas.MatchingAlgorithm.Data/Repositories/DonorRetrieval/DonorSearchRepository.cs` line 200 at r4](https://reviewable.io/reviews/Anthony-Nolan/Atlas/1286#-O12GXBZCOD46K4iT1aT:-O12GXBZCOD46K4iT1aU:b-bvjqrj) ([raw file](https://github.com/Anthony-Nolan/Atlas/blob/921e7451bd3ee72b7d2ce5db19a2d3e441c96abf/Atlas.MatchingAlgorithm.Data/Repositories/DonorRetrieval/DonorSearchRepository.cs#L200)):*
> ```Smalltalk
> {donorIdTempTableJoin}
> {donorUpdatedJoin}
> {donorTypeFilteredJoin}
> ```

Can we make a single join for both filtering on DonorType and RegistryCode? This way we'll make use of the index by DonorType, RegistryCode

daria-sorokina-da (commented):
> *[Reviewable](https://reviewable.io/reviews/Anthony-Nolan/Atlas/1286)* status: all files reviewed, 9 unresolved discussions (waiting on @SergeyEzh and @zabeen)

*[a discussion](https://reviewable.io/reviews/Anthony-Nolan/Atlas/1286#-O12QVzR9DTJoUqWu-J1:-O12QVzR9DTJoUqWu-J2:b-y2bhoq) (no related file):*
There's a merge conflict - please rebase onto rc/2.3.0

___

SergeyEzh (commented):
> *[Reviewable](https://reviewable.io/reviews/Anthony-Nolan/Atlas/1286)* status: all files reviewed, 9 unresolved discussions (waiting on @daria-sorokina-da and @zabeen)

___
*[`Atlas.MatchingAlgorithm.Data/Migrations/20240703115406_AddDonorTypeIndexToDonorsTable.cs` line 12 at r4](https://reviewable.io/reviews/Anthony-Nolan/Atlas/1286#-O12APup2hCjqzVPeFvB:-O1HoJ3i4b7OUiaIww-N:b3flhy1) ([raw file](https://github.com/Anthony-Nolan/Atlas/blob/921e7451bd3ee72b7d2ce5db19a2d3e441c96abf/Atlas.MatchingAlgorithm.Data/Migrations/20240703115406_AddDonorTypeIndexToDonorsTable.cs#L12)):*

Let's change this migration to first check whether the index is not there yet, and only if it's not yet there - add it.
I think we'll have to create it ahead of time as it may take a while to create
</blockquote></details>

hmm, I don't think that it is possible with MigrationBuilder api. It can be achieved by creating index by running custom sql. Definitely we can do this, but why not apply the migration ahead of time instead?
___
*[`Atlas.MatchingAlgorithm.Data/Migrations/20240703115406_AddDonorTypeIndexToDonorsTable.cs` line 15 at r4](https://reviewable.io/reviews/Anthony-Nolan/Atlas/1286#-O12AqVEElvrSUHUWGUW:-O1HpaF124v7VR-PaNcb:b-juwd2x) ([raw file](https://github.com/Anthony-Nolan/Atlas/blob/921e7451bd3ee72b7d2ce5db19a2d3e441c96abf/Atlas.MatchingAlgorithm.Data/Migrations/20240703115406_AddDonorTypeIndexToDonorsTable.cs#L15)):*

Shouldn't the index be on two columns - DonorType and RegistryCode?
Making an index only by DonorType is not very helpful, because there are just two values in DonorType - so it's like indexing on a boolean value. It would speed up queries for Cords, i.e. if there are much less cords than adult donors.
DonorType will be less selective than RegistryCode, but it will always be provided in the query.

I'd suggest to have an index on (DonorType, RegistryCode) include (DonorId) - then we'll cover both scenarios
</blockquote></details>

this index was suggested by sql server when I ran queries on dev environment. Today checking your suggestion, I realized that sql server didn't consider (DonorType, RegistryCode) index as option because RegistryCode has type incompatible with indexes. Currently it is nvarchar(max). We need limit its length with some reasonable value. 64 characters may be? I tested with amended column type and (DonorType, RegistryCode) index and it work's faster.
___
*[`Atlas.MatchingAlgorithm.Data/Migrations/20240703115406_AddDonorTypeIndexToDonorsTable.cs` line 16 at r4](https://reviewable.io/reviews/Anthony-Nolan/Atlas/1286#-O12BKqf-PnyhSachwBa:-O1HqzdU61zkL6cfhjzA:bm36z0q) ([raw file](https://github.com/Anthony-Nolan/Atlas/blob/921e7451bd3ee72b7d2ce5db19a2d3e441c96abf/Atlas.MatchingAlgorithm.Data/Migrations/20240703115406_AddDonorTypeIndexToDonorsTable.cs#L16)):*

We're only returning DonorId by this filter, right? That's why we don't need to include other columns?
</blockquote></details>

RegistryCode is included in index to avoid additional primary key lookup after index lookup to get RegistryCode value
___
*[`Atlas.MatchingAlgorithm.Data/Repositories/DonorRetrieval/DonorSearchRepository.cs` line 146 at r4](https://reviewable.io/reviews/Anthony-Nolan/Atlas/1286#-O12C9DZ5yakWMRJacv5:-O1Hs6oMDtq7a6jcxFZF:bf31y96) ([raw file](https://github.com/Anthony-Nolan/Atlas/blob/921e7451bd3ee72b7d2ce5db19a2d3e441c96abf/Atlas.MatchingAlgorithm.Data/Repositories/DonorRetrieval/DonorSearchRepository.cs#L146)):*

Do we still need selectDonorIdStatement variable?
</blockquote></details>

yes. It is also used in last query after `SELECT DISTINCT `
___
*[`Atlas.MatchingAlgorithm.Data/Repositories/DonorRetrieval/DonorSearchRepository.cs` line 200 at r4](https://reviewable.io/reviews/Anthony-Nolan/Atlas/1286#-O12GXBZCOD46K4iT1aT:-O1HsHxr6WiiOkd_xMe3:bioir2a) ([raw file](https://github.com/Anthony-Nolan/Atlas/blob/921e7451bd3ee72b7d2ce5db19a2d3e441c96abf/Atlas.MatchingAlgorithm.Data/Repositories/DonorRetrieval/DonorSearchRepository.cs#L200)):*

Can we make a single join for both filtering on DonorType and RegistryCode? This way we'll make use of the index by DonorType, RegistryCode
</blockquote></details>

Do you mean to write something like this `INNER JOIN Donors d ON m.DonorId = d.DonorId AND d.DonorType = {(int)filteringOptions.DonorType}" and d.RegistryCode in (...)` ?

With creating index by DonorType and RegistryCode, it will be used even without changes in this query. But proposed variant will work faster if users will run queries with only few registry codes. Because overhead for filling temp table would be to big. I think your suggestion make sense if there's no more than 20-30 registry codes in search request.

daria-sorokina-da (commented):
> *[Reviewable](https://reviewable.io/reviews/Anthony-Nolan/Atlas/1286)* status: all files reviewed, 8 unresolved discussions (waiting on @SergeyEzh and @zabeen)

___
*[`Atlas.MatchingAlgorithm.Data/Migrations/20240703115406_AddDonorTypeIndexToDonorsTable.cs` line 12 at r4](https://reviewable.io/reviews/Anthony-Nolan/Atlas/1286#-O12APup2hCjqzVPeFvB:-O1Hw5C2-Tchq44WP6MM:b-1cjlf6) ([raw file](https://github.com/Anthony-Nolan/Atlas/blob/921e7451bd3ee72b7d2ce5db19a2d3e441c96abf/Atlas.MatchingAlgorithm.Data/Migrations/20240703115406_AddDonorTypeIndexToDonorsTable.cs#L12)):*
<details><summary><i>Previously, SergeyEzh wrote…</i></summary><blockquote>

hmm, I don't think that it is possible with MigrationBuilder api. It can be achieved by creating index by running custom sql. Definitely we can do this, but why not apply the migration ahead of time instead?
</blockquote></details>

Yes, let's create the index using a custom SQL.

Migration is part of our release pipeline, so it should be run during the release.

I prefer manual application in this case because this index can take some time to create, and we'll need to run it on an online database, without locking the data. We'll need to figure out the best way to do that on a database copy first in the Live WMDA environment. All this said - it would be much easier to do all these experiments from SQL script than from a migration.
___
*[`Atlas.MatchingAlgorithm.Data/Migrations/20240703115406_AddDonorTypeIndexToDonorsTable.cs` line 15 at r4](https://reviewable.io/reviews/Anthony-Nolan/Atlas/1286#-O12AqVEElvrSUHUWGUW:-O1HyVj0C0-MFmsjpwYA:bszq9j) ([raw file](https://github.com/Anthony-Nolan/Atlas/blob/921e7451bd3ee72b7d2ce5db19a2d3e441c96abf/Atlas.MatchingAlgorithm.Data/Migrations/20240703115406_AddDonorTypeIndexToDonorsTable.cs#L15)):*
<details><summary><i>Previously, SergeyEzh wrote…</i></summary><blockquote>

this index was suggested by sql server when I ran queries on dev environment. Today checking your suggestion, I realized that sql server didn't consider (DonorType, RegistryCode) index as option because RegistryCode has type incompatible with indexes. Currently it is nvarchar(max). We need limit its length with some reasonable value. 64 characters may be? I tested with amended column type and (DonorType, RegistryCode) index and it work's faster.
</blockquote></details>

In Donors.Donors it has type nvarchar(256), let's keep it the same?
Please change EthnicityCode as well - also to nvarchar(256).

Let's update of the column type, but same as the index migration - please make this a custom SQL.
Check if the column is not yet of the target type - and change only in that case.

We'll have to run some checks and potentially do the type update ahead of time as this may take time on the WMDA Live Database.
___
*[`Atlas.MatchingAlgorithm.Data/Migrations/20240703115406_AddDonorTypeIndexToDonorsTable.cs` line 16 at r4](https://reviewable.io/reviews/Anthony-Nolan/Atlas/1286#-O12BKqf-PnyhSachwBa:-O1Hzp328Z-sCC-Xy1q4:b53k6k4) ([raw file](https://github.com/Anthony-Nolan/Atlas/blob/921e7451bd3ee72b7d2ce5db19a2d3e441c96abf/Atlas.MatchingAlgorithm.Data/Migrations/20240703115406_AddDonorTypeIndexToDonorsTable.cs#L16)):*
<details><summary><i>Previously, SergeyEzh wrote…</i></summary><blockquote>

RegistryCode is included in index to avoid additional primary key lookup after index lookup to get RegistryCode value
</blockquote></details>

If we get rid of the temp table and use the in clause in the same join statement as filtering by donor type, then maybe we don't need to include it?
___
*[`Atlas.MatchingAlgorithm.Data/Repositories/DonorRetrieval/DonorSearchRepository.cs` line 200 at r4](https://reviewable.io/reviews/Anthony-Nolan/Atlas/1286#-O12GXBZCOD46K4iT1aT:-O1I-Dec4QyLKS4ESz1j:bm7skz5) ([raw file](https://github.com/Anthony-Nolan/Atlas/blob/921e7451bd3ee72b7d2ce5db19a2d3e441c96abf/Atlas.MatchingAlgorithm.Data/Repositories/DonorRetrieval/DonorSearchRepository.cs#L200)):*
<details><summary><i>Previously, SergeyEzh wrote…</i></summary><blockquote>

Do you mean to write something like this `INNER JOIN Donors d ON m.DonorId = d.DonorId AND d.DonorType = {(int)filteringOptions.DonorType}" and d.RegistryCode in (...)` ?

With creating index by DonorType and RegistryCode, it will be used even without changes in this query. But proposed variant will work faster if users will run queries with only few registry codes. Because overhead for filling temp table would be to big. I think your suggestion make sense if there's no more than 20-30 registry codes in search request.
</blockquote></details>

Currently the number of registry codes are 1-5, and overall the maximum number is 99 (that's how many registries there are in the world).
So I think we can get rid of the temp table and just use the in clause.

daria-sorokina-da (commented):
> *[Reviewable](https://reviewable.io/reviews/Anthony-Nolan/Atlas/1286)* status: all files reviewed, 8 unresolved discussions (waiting on @SergeyEzh and @zabeen)

___
*[`Atlas.MatchingAlgorithm.Data/Repositories/DonorRetrieval/DonorSearchRepository.cs` line 200 at r4](https://reviewable.io/reviews/Anthony-Nolan/Atlas/1286#-O12GXBZCOD46K4iT1aT:-O1I130k6LFR9hip43mm:bvl4uh2) ([raw file](https://github.com/Anthony-Nolan/Atlas/blob/921e7451bd3ee72b7d2ce5db19a2d3e441c96abf/Atlas.MatchingAlgorithm.Data/Repositories/DonorRetrieval/DonorSearchRepository.cs#L200)):*

Currently the number of registry codes are 1-5, and overall the maximum number is 99 (that's how many registries there are in the world).
So I think we can get rid of the temp table and just use the in clause.
</blockquote></details>

If you're curious - here are all the registry codes - https://statistics.wmda.info/

SergeyEzh (commented):
> *[Reviewable](https://reviewable.io/reviews/Anthony-Nolan/Atlas/1286)* status: 22 of 31 files reviewed, 8 unresolved discussions (waiting on @daria-sorokina-da and @zabeen)

*[a discussion](https://reviewable.io/reviews/Anthony-Nolan/Atlas/1286#-O12QVzR9DTJoUqWu-J1:-O1MzzjjAgvtX5jbmLb2:bb7f2mf) (no related file):*

There's a merge conflict - please rebase onto rc/2.3.0
</blockquote></details>

done

___
___
*[`Atlas.MatchingAlgorithm/Services/Search/Matching/PerLocusDonorMatchingService.cs` line 69 at r4](https://reviewable.io/reviews/Anthony-Nolan/Atlas/1286#-O12A7qx3P9dkbTObvYi:-O1N-1rJ4WW0Okwtr94H:b-hfpv0l) ([raw file](https://github.com/Anthony-Nolan/Atlas/blob/921e7451bd3ee72b7d2ce5db19a2d3e441c96abf/Atlas.MatchingAlgorithm/Services/Search/Matching/PerLocusDonorMatchingService.cs#L69)):*

do we need the comment?
</blockquote></details>

removed
___
*[`Atlas.MatchingAlgorithm.Data/Repositories/DonorRetrieval/DonorSearchRepository.cs` line 200 at r4](https://reviewable.io/reviews/Anthony-Nolan/Atlas/1286#-O12GXBZCOD46K4iT1aT:-O1N-Bcx4HqAvi88o835:b-hdoan1) ([raw file](https://github.com/Anthony-Nolan/Atlas/blob/921e7451bd3ee72b7d2ce5db19a2d3e441c96abf/Atlas.MatchingAlgorithm.Data/Repositories/DonorRetrieval/DonorSearchRepository.cs#L200)):*

If you're curious - here are all the registry codes - https://statistics.wmda.info/
</blockquote></details>

changed to use `IN` for RegistryCode filtering
___
*[`Atlas.MatchingAlgorithm.Data/Migrations/20240703115406_AddDonorTypeIndexToDonorsTable.cs` line 12 at r4](https://reviewable.io/reviews/Anthony-Nolan/Atlas/1286#-O12APup2hCjqzVPeFvB:-O1N-Q3S0L65lJcH_os9:b-nm2tt0) ([raw file](https://github.com/Anthony-Nolan/Atlas/blob/921e7451bd3ee72b7d2ce5db19a2d3e441c96abf/Atlas.MatchingAlgorithm.Data/Migrations/20240703115406_AddDonorTypeIndexToDonorsTable.cs#L12)):*

Yes, let's create the index using a custom SQL.

Migration is part of our release pipeline, so it should be run during the release.

I prefer manual application in this case because this index can take some time to create, and we'll need to run it on an online database, without locking the data. We'll need to figure out the best way to do that on a database copy first in the Live WMDA environment. All this said - it would be much easier to do all these experiments from SQL script than from a migration.
</blockquote></details>

created custom sql
___
*[`Atlas.MatchingAlgorithm.Data/Migrations/20240703115406_AddDonorTypeIndexToDonorsTable.cs` line 13 at r4](https://reviewable.io/reviews/Anthony-Nolan/Atlas/1286#-O12Aafp4GtmmN3Lg1cg:-O1N5V625RDTKiH39N4Y:b-896fix) ([raw file](https://github.com/Anthony-Nolan/Atlas/blob/921e7451bd3ee72b7d2ce5db19a2d3e441c96abf/Atlas.MatchingAlgorithm.Data/Migrations/20240703115406_AddDonorTypeIndexToDonorsTable.cs#L13)):*

The name should IX_Donors_DonorType_RegistryCode, I think
</blockquote></details>

Done.
___
*[`Atlas.MatchingAlgorithm.Data/Migrations/20240703115406_AddDonorTypeIndexToDonorsTable.cs` line 15 at r4](https://reviewable.io/reviews/Anthony-Nolan/Atlas/1286#-O12AqVEElvrSUHUWGUW:-O1N5dmy2R_qJ50PuGeR:b-896fix) ([raw file](https://github.com/Anthony-Nolan/Atlas/blob/921e7451bd3ee72b7d2ce5db19a2d3e441c96abf/Atlas.MatchingAlgorithm.Data/Migrations/20240703115406_AddDonorTypeIndexToDonorsTable.cs#L15)):*

In Donors.Donors it has type nvarchar(256), let's keep it the same?
Please change EthnicityCode as well - also to nvarchar(256).

Let's update of the column type, but same as the index migration - please make this a custom SQL.
Check if the column is not yet of the target type - and change only in that case.

We'll have to run some checks and potentially do the type update ahead of time as this may take time on the WMDA Live Database.
</blockquote></details>

Done.
___
*[`Atlas.MatchingAlgorithm.Data/Migrations/20240703115406_AddDonorTypeIndexToDonorsTable.cs` line 16 at r4](https://reviewable.io/reviews/Anthony-Nolan/Atlas/1286#-O12BKqf-PnyhSachwBa:-O1N5kNk9qF0kcTOnahl:bq83waq) ([raw file](https://github.com/Anthony-Nolan/Atlas/blob/921e7451bd3ee72b7d2ce5db19a2d3e441c96abf/Atlas.MatchingAlgorithm.Data/Migrations/20240703115406_AddDonorTypeIndexToDonorsTable.cs#L16)):*

If we get rid of the temp table and use the in clause in the same join statement as filtering by donor type, then maybe we don't need to include it?
</blockquote></details>

I'd say using in clause isn't changing anything in that case. But since we're using another index there's nothing to worry about.

daria-sorokina-da (commented):
> Reviewed 9 of 9 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/Anthony-Nolan/Atlas/1286)* status: all files reviewed, 2 unresolved discussions (waiting on @SergeyEzh and @zabeen)

*[a discussion](https://reviewable.io/reviews/Anthony-Nolan/Atlas/1286#-O1NA96S5t22JVVpH5Fv:-O1NA96S5t22JVVpH5Fw:btyrwps) (no related file):*
Looks good to me. We will merge it once we create the index and update the columns in DEV

___

daria-sorokina-da (commented):
> *[Reviewable](https://reviewable.io/reviews/Anthony-Nolan/Atlas/1286)* status: all files reviewed, 3 unresolved discussions (waiting on @SergeyEzh and @zabeen)

___
*[`Atlas.MatchingAlgorithm.Data/Migrations/20240709124454_AddDonorTypeAndRegistryCodeIndex.cs` line 15 at r5](https://reviewable.io/reviews/Anthony-Nolan/Atlas/1286#-O3DMK_I2qpO6lScVr2v:-O3DMK_I2qpO6lScVr2w:bkpzhiq) ([raw file](https://github.com/Anthony-Nolan/Atlas/blob/137d79a5c9870abc38f4cc0a2ae7dfad58c89668/Atlas.MatchingAlgorithm.Data/Migrations/20240709124454_AddDonorTypeAndRegistryCodeIndex.cs#L15)):*
> ```Smalltalk
> 
>         private const string changeColumnsTypeSql = @"
>             ALTER TABLE [Donors] ALTER COLUMN [EthnicityCode] nvarchar(256) NULL;
> ```

Please add a check that the column is not yet of this type before altering

daria-sorokina-da (commented):
> *[Reviewable](https://reviewable.io/reviews/Anthony-Nolan/Atlas/1286)* status: all files reviewed, 4 unresolved discussions (waiting on @SergeyEzh and @zabeen)

___
*[`Atlas.MatchingAlgorithm.Data/Migrations/20240709124454_AddDonorTypeAndRegistryCodeIndex.cs` line 11 at r5](https://reviewable.io/reviews/Anthony-Nolan/Atlas/1286#-O3DMook6YidQBnizncl:-O3DMook6YidQBnizncm:bxx3p0c) ([raw file](https://github.com/Anthony-Nolan/Atlas/blob/137d79a5c9870abc38f4cc0a2ae7dfad58c89668/Atlas.MatchingAlgorithm.Data/Migrations/20240709124454_AddDonorTypeAndRegistryCodeIndex.cs#L11)):*
> ```Smalltalk
>         private const string createIndexSql = @"
>             if not exists (SELECT 1 FROM sys.indexes WHERE name = 'IX_Donors_DonorType_RegistryCode' AND object_id = OBJECT_ID('dbo.Donors')) 
> 	            CREATE INDEX [IX_Donors_DonorType_RegistryCode] ON [Donors] ([DonorType], [RegistryCode]);
> ```

Please add keyword WITH (ONLINE = ON) so that the table does not get locked while the index is being created

daria-sorokina-da (commented):
> *[Reviewable](https://reviewable.io/reviews/Anthony-Nolan/Atlas/1286)* status: all files reviewed, 3 unresolved discussions (waiting on @SergeyEzh and @zabeen)

___
*[`Atlas.MatchingAlgorithm.Data/Migrations/20240709124454_AddDonorTypeAndRegistryCodeIndex.cs` line 15 at r5](https://reviewable.io/reviews/Anthony-Nolan/Atlas/1286#-O3DMK_I2qpO6lScVr2v:-O3Dc0sxCs1LPJ1Qepa1:b-sy262m) ([raw file](https://github.com/Anthony-Nolan/Atlas/blob/137d79a5c9870abc38f4cc0a2ae7dfad58c89668/Atlas.MatchingAlgorithm.Data/Migrations/20240709124454_AddDonorTypeAndRegistryCodeIndex.cs#L15)):*

Please add a check that the column is not yet of this type before altering
</blockquote></details>

Just tested this - and I think we're OK as is - the 2nd time this query is processed quickly (so it does nothing)

SergeyEzh (commented):
git 

> *[Reviewable](https://reviewable.io/reviews/Anthony-Nolan/Atlas/1286#-:-O3XHpEJ6ltMtwcirBCM:bb7gv83)* status: all files reviewed, 3 unresolved discussions (waiting on @daria-sorokina-da and @zabeen)

*[a discussion](https://reviewable.io/reviews/Anthony-Nolan/Atlas/1286#-O1NA96S5t22JVVpH5Fv:-O3XK1tq7SjpUNvfS5NP:b44tpx) (no related file):*

Looks good to me. We will merge it once we create the index and update the columns in DEV
</blockquote></details>

👍

___
___
*[`Atlas.MatchingAlgorithm.Data/Migrations/20240709124454_AddDonorTypeAndRegistryCodeIndex.cs` line 11 at r5](https://reviewable.io/reviews/Anthony-Nolan/Atlas/1286#-O3DMook6YidQBnizncl:-O3XJzn_24OAPvk81Qpp:bb7f2mf) ([raw file](https://github.com/Anthony-Nolan/Atlas/blob/137d79a5c9870abc38f4cc0a2ae7dfad58c89668/Atlas.MatchingAlgorithm.Data/Migrations/20240709124454_AddDonorTypeAndRegistryCodeIndex.cs#L11)):*

Please add keyword WITH (ONLINE = ON) so that the table does not get locked while the index is being created
</blockquote></details>

done

daria-sorokina-da (commented):
> Reviewed 1 of 1 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/Anthony-Nolan/Atlas/1286)* status: all files reviewed, 2 unresolved discussions (waiting on @SergeyEzh and @zabeen)

*[a discussion](https://reviewable.io/reviews/Anthony-Nolan/Atlas/1286#-O1NA96S5t22JVVpH5Fv:-O3Y5cLf4bsjxsG1v_JB:b-s59dx) (no related file):*
<details><summary><i>Previously, SergeyEzh wrote…</i></summary><blockquote>

👍
</blockquote></details>

Index and column updates were done in DEV, old code continues working with these database changes.
We're now in progress with getting some stats before mering this PR

___

PR #1324, opened 2024-07-01
fix #1318: donor dto mapping fix

### LINKED ISSUE #1318: Data Refresh - Change the way LastUpdateDateTime column in the DonorManagementLogs table (A/B databases) is populated during data reshresh
### Context
Currently, during Data Refresh, the LastUpdateDateTime column in the DonorManagementLogs in transient (A/B) databases is set to the date and time of Date Refresh processing.

Example: in the WMDA Live environment the last Data Refresh happened on 2024-04-26, and in the DonorManagementLogs we see lots of records with LastUpdateDateTime 2024-04-26, see:
image

Here "-1" in the SequenceNumberOfLastUpdate indicates that the Data Refresh process created this record.

### Issue description
This approach causes issues during Repeat Searches run after the data refresh.
That is because all the donors added during Data Refresh now have LastUpdateDateTime after the last cut-off date of the search.
So basically Repeat Searches perform the search from scratch (return all the same donors as the original search, as if they changed).

This causes performance issues on Atlas side and increases the costs in Azure, and causes performance issues on the client side as well, as the client does not expect the Repeat Searches to return lots of results.

### Proposed fix
We need to start filling the LastUpdateDateTime with the date when the donor was last updated, instead of the Data Refresh date.
We should take the date from the **persistent database Donors.Donors table - LastUpdated**. 

### Solution details

1.  In `Atlas.DonorImport.Data.Repositories` `DonorReadRepository.StreamAllDonors` update the returned model - add LastUpdated date to it.
2. Start using the LastUpdated in `Atlas.MatchingAlgorithm.Services.DataRefresh.DonorImport` `DonorImporter.InsertDonorBatch` - instead of `UpdateDateTime = batchFetchTime` we should now have `UpdateDateTime = d.LastUpdated`

SergeyEzh:
This change is [Reviewable](https://reviewable.io/reviews/Anthony-Nolan/Atlas/1324)

### DIFF
--- a/Atlas.DonorImport/Models/Mapping/DatabaseDonorMaps.cs
+++ b/Atlas.DonorImport/Models/Mapping/DatabaseDonorMaps.cs
@@ -13,6 +13,7 @@ namespace Atlas.DonorImport.Models.Mapping
                 DonorType = donor.DonorType.ToMatchingAlgorithmType(),
                 EthnicityCode = donor.EthnicityCode,
                 RegistryCode = donor.RegistryCode,
+                LastUpdated = donor.LastUpdated,
                 A_1 = donor.A_1,
                 A_2 = donor.A_2,
                 B_1 = donor.B_1,


daria-sorokina-da (approved):
> Reviewed 1 of 1 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/Anthony-Nolan/Atlas/1324)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @pchenery)

PR #1358, opened 2024-08-14
feature: CBT-1476: Send Matching Algorithm core matching started event

pchenery:
This change is [Reviewable](https://reviewable.io/reviews/Anthony-Nolan/Atlas/1358)

### DIFF
--- a/Atlas.MatchingAlgorithm.Functions/Functions/SearchFunctions.cs
+++ b/Atlas.MatchingAlgorithm.Functions/Functions/SearchFunctions.cs
@@ -58,6 +58,7 @@ namespace Atlas.MatchingAlgorithm.Functions.Functions
             int deliveryCount,
             DateTime enqueuedTimeUtc)
         {
+            enqueuedTimeUtc = DateTime.SpecifyKind(enqueuedTimeUtc, DateTimeKind.Utc);
             await searchRunner.RunSearch(request, deliveryCount, enqueuedTimeUtc);
         }
 
--- a/Atlas.MatchingAlgorithm.Test/Services/Search/MatchingAlgorithmSearchTrackingDispatcherTests.cs
+++ b/Atlas.MatchingAlgorithm.Test/Services/Search/MatchingAlgorithmSearchTrackingDispatcherTests.cs
@@ -55,5 +55,35 @@ namespace Atlas.MatchingAlgorithm.Test.Services.Search
 
             actualAttemptStartedEvent.Should().BeEquivalentTo(expectedMatchingAlgorithmAttemptStartedEvent);
         }
+
+        [TestCase(SearchTrackingEventType.MatchingAlgorithmPersistingResultsEnded)]
+        [TestCase(SearchTrackingEventType.MatchingAlgorithmPersistingResultsStarted)]
+        [TestCase(SearchTrackingEventType.MatchingAlgorithmCoreScoringEnded)]
+        [TestCase(SearchTrackingEventType.MatchingAlgorithmCoreScoringStarted)]
+        [TestCase(SearchTrackingEventType.MatchingAlgorithmCoreMatchingEnded)]
+        [TestCase(SearchTrackingEventType.MatchingAlgorithmCoreMatchingStarted)]
+        [Test]
+        public async Task DispatchMatchingAlgorithmAttemptTimingEvent_WhenTimingEventReceived_DispatchesEvent(SearchTrackingEventType eventType)
+        {
+            const byte attemptNumber = 1;
+            var eventTime = new DateTime(2024, 8, 13);
+
+            var expectedMatchingAlgorithmAttemptTimingEvent = new MatchingAlgorithmAttemptTimingEvent()
+            {
+                SearchRequestId = new Guid("aaaaaaaa-bbbb-cccc-dddd-eeeeeeeeeeee"),
+                AttemptNumber = attemptNumber,
+                TimeUtc = eventTime
+            };
+
+            MatchingAlgorithmAttemptTimingEvent actualAttemptTimingEvent = null;
+
+            await searchTrackingServiceBusClient.PublishSearchTrackingEvent(
+                Arg.Do<MatchingAlgorithmAttemptTimingEvent>(x => actualAttemptTimingEvent = x),
+                Arg.Is(eventType));
+
+            await searchTrackingDispatcher.DispatchMatchingAlgorithmAttemptTimingEvent(eventType, eventTime);
+
+            actualAttemptTimingEvent.Should().BeEquivalentTo(expectedMatchingAlgorithmAttemptTimingEvent);
+        }
     }
 }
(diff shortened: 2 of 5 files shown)


daria-sorokina-da (changes requested):
> Reviewed 3 of 3 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/Anthony-Nolan/Atlas/1358)* status: all files reviewed, 1 unresolved discussion (waiting on @pchenery and @SergeyEzh)

___
*[`Atlas.MatchingAlgorithm/Services/Search/SearchRunner.cs` line 95 at r1](https://reviewable.io/reviews/Anthony-Nolan/Atlas/1358#-O4GkYRXCqhBP3v1LhJz:-O4GkYRXCqhBP3v1LhK-:b-9nsng7) ([raw file](https://github.com/Anthony-Nolan/Atlas/blob/30ca816e9a3bf5d5746e964cc5ce48b95befe181/Atlas.MatchingAlgorithm/Services/Search/SearchRunner.cs#L95)):*
> ```Smalltalk
>                 matchingAlgorithmSearchTrackingContextManager.Set(context);
>                 await matchingAlgorithmSearchTrackingDispatcher.DispatchInitiationEvent(enqueuedTimeUtc.LocalDateTime, searchStartTime.UtcDateTime);
>                 await matchingAlgorithmSearchTrackingDispatcher.DispatchMatchingAlgorithmAttemptTimingEvent(
> ```

Did it start here though? I think we need to add this code into MatchingService.GetMatches - because this is where we actually do the matching.

But we need to be clever there because there can be multiple sub-searches.
I think we're good to just send an event for each sub-search because:
1) all sub-searches will start at the same time, so even if the date gets updated by the next call, we're good, they'll be close to each other
2) sub-searches will be finishing at different times, but here we're also good as we want to know the LATEST date, so the value will be overridden by the latest date, and that's OK

daria-sorokina-da (changes requested):
> *[Reviewable](https://reviewable.io/reviews/Anthony-Nolan/Atlas/1358)* status: all files reviewed, 2 unresolved discussions (waiting on @pchenery and @SergeyEzh)

___
*[`Atlas.MatchingAlgorithm/Services/Search/SearchRunner.cs` line 94 at r1](https://reviewable.io/reviews/Anthony-Nolan/Atlas/1358#-O4GnYjcCd9regme2BGY:-O4GnYjcCd9regme2BGZ:b-gfmzje) ([raw file](https://github.com/Anthony-Nolan/Atlas/blob/30ca816e9a3bf5d5746e964cc5ce48b95befe181/Atlas.MatchingAlgorithm/Services/Search/SearchRunner.cs#L94)):*
> ```Smalltalk
> 
>                 matchingAlgorithmSearchTrackingContextManager.Set(context);
>                 await matchingAlgorithmSearchTrackingDispatcher.DispatchInitiationEvent(enqueuedTimeUtc.LocalDateTime, searchStartTime.UtcDateTime);
> ```

Don't think changing local to UTC is correct though as both dates should be in UTC. Why do we need this change?

daria-sorokina-da (changes requested):
> *[Reviewable](https://reviewable.io/reviews/Anthony-Nolan/Atlas/1358)* status: all files reviewed, 2 unresolved discussions (waiting on @pchenery and @SergeyEzh)

___
*[`Atlas.MatchingAlgorithm/Services/Search/SearchRunner.cs` line 94 at r1](https://reviewable.io/reviews/Anthony-Nolan/Atlas/1358#-O4GnYjcCd9regme2BGY:-O4GoS8x1hG0dXZOImCi:bu5dbf4) ([raw file](https://github.com/Anthony-Nolan/Atlas/blob/30ca816e9a3bf5d5746e964cc5ce48b95befe181/Atlas.MatchingAlgorithm/Services/Search/SearchRunner.cs#L94)):*

Don't think changing local to UTC is correct though as both dates should be in UTC. Why do we need this change?
</blockquote></details>

Sorry, my message is incorrect, correction below:
* Don't think changing UTC to local is correct though as both dates should be in UTC. Why do we need this change?

pchenery (commented):
> *[Reviewable](https://reviewable.io/reviews/Anthony-Nolan/Atlas/1358)* status: 2 of 5 files reviewed, 2 unresolved discussions (waiting on @daria-sorokina-da and @SergeyEzh)

___
*[`Atlas.MatchingAlgorithm/Services/Search/SearchRunner.cs` line 94 at r1](https://reviewable.io/reviews/Anthony-Nolan/Atlas/1358#-O4GnYjcCd9regme2BGY:-O4K72de7Tp47RGx0KsU:b-896fix) ([raw file](https://github.com/Anthony-Nolan/Atlas/blob/30ca816e9a3bf5d5746e964cc5ce48b95befe181/Atlas.MatchingAlgorithm/Services/Search/SearchRunner.cs#L94)):*

I think the problem is that in 
```
[SuppressMessage(null, SuppressMessage.UnusedParameter, Justification = SuppressMessage.UsedByAzureTrigger)]
[Function(nameof(RunSearch))]
public async Task RunSearch(
    [ServiceBusTrigger(
        "%MessagingServiceBus:SearchRequestsTopic%",
        "%MessagingServiceBus:SearchRequestsSubscription%",
        Connection = "MessagingServiceBus:ConnectionString")]
    IdentifiedSearchRequest request,
    int deliveryCount,
    DateTime enqueuedTimeUtc)
{
    await searchRunner.RunSearch(request, deliveryCount, enqueuedTimeUtc);
}
```
an implicit conversion happens between enqueuedTimeUtc (which is a DateTime) and DateTimeOffset and smth is wrong there.

Probably we need to specify the date time kind for this date as UTC for the conversion to happen correctly.

```
enqueuedTimeUtc= DateTime.SpecifyKind(enqueuedTimeUtc, DateTimeKind.Utc);
```
</blockquote></details>

Done.
___
*[`Atlas.MatchingAlgorithm/Services/Search/SearchRunner.cs` line 95 at r1](https://reviewable.io/reviews/Anthony-Nolan/Atlas/1358#-O4GkYRXCqhBP3v1LhJz:-O4Kdnrb1Wq1j7cJFZm_:b-896fix) ([raw file](https://github.com/Anthony-Nolan/Atlas/blob/30ca816e9a3bf5d5746e964cc5ce48b95befe181/Atlas.MatchingAlgorithm/Services/Search/SearchRunner.cs#L95)):*

Did it start here though? I think we need to add this code into MatchingService.GetMatches - because this is where we actually do the matching.

But we need to be clever there because there can be multiple sub-searches.
I think we're good to just send an event for each sub-search because:
1) all sub-searches will start at the same time, so even if the date gets updated by the next call, we're good, they'll be close to each other
2) sub-searches will be finishing at different times, but here we're also good as we want to know the LATEST date, so the value will be overridden by the latest date, and that's OK
</blockquote></details>

Done.

daria-sorokina-da (approved):
> Reviewed 3 of 3 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/Anthony-Nolan/Atlas/1358)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @SergeyEzh)

PR #1479, opened 2026-07-14
docs: ATL-187: add CLAUDE.md with build/test/architecture guidance

da-ikovalenko:
## Summary
- Adds `CLAUDE.md` to give Claude Code (and future contributors) guidance on build/test commands, architecture, DI/messaging/settings conventions, and branching rules for this repo.

## Test plan
- Docs-only change; no build/test impact.

- - -
This change is [Reviewable](https://reviewable.io/reviews/Anthony-Nolan/Atlas/1479)

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -24,7 +24,8 @@ dotnet restore
 dotnet build --configuration Release --no-restore
 
 # Run all unit tests (Test projects only, excludes Integration/Validation/Performance/Verification)
-dotnet test "**/*Test.csproj"
+shopt -s globstar
+dotnet test **/*Test.csproj
 
 # Run a single test project
 dotnet test Atlas.MatchingAlgorithm.Test/Atlas.MatchingAlgorithm.Test.csproj
@@ -47,7 +48,7 @@ Test projects follow a strict per-component split, each with different infra req
 - `*.Test.Validation` (Matching, Match Prediction) — spins up a real in-memory server and exercises it over HTTP; written in Gherkin via Reqnroll for non-technical review by the Search/BioInformatics teams; requires user secrets and a specific HLA Metadata Dictionary version to be generated locally first.
 - `*.Test.Performance` / `*.Test.Verification` — manual/CI-nightly performance harnesses, gated by the custom `IgnoreExceptOnCiPerfTestAttribute` (only runs when `RUN_CI_PERF_TESTS` env var is set).
 
-There is no local Azure Service Bus emulator — local dev requires a real Azure Service Bus namespace with manually-created topics/subscriptions (see `README_DevelopmentStartUpGuide.md` for the exact topic/subscription names needed per workflow).
+There is no local Azure Service Bus emulator in use for this project — local dev requires a real Azure Service Bus namespace with manually-created topics/subscriptions (see `README_DevelopmentStartUpGuide.md` for the exact topic/subscription names needed per workflow). Microsoft does ship a dockerized Service Bus emulator, but it isn't adopted here (it caps out at ~10 concurrent requests and is clunky to configure).
 
 ## Architecture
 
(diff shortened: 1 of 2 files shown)


CLAUDE.md
```diff
@@ -0,0 +1,102 @@
...
+- **Feature/fix branches**: branch from the `rc/` or `hotfix/` branch that the change is targeting, and PR back into that same `rc/`/`hotfix/` branch — never into `master` directly.
+- `master` is only advanced by the release/hotfix process itself (e.g. `post_release_candidate_branch_creation.yml` bumps the version on `master` when a new `rc/*` branch is cut), not by regular feature work.
+
+When creating a branch or opening a PR, confirm which `rc/`/`hotfix/` branch is the intended target before assuming `master`.
+
+> **Inconsistency found in existing docs:** `README_Contribution_Versioning.md` (Contributing section) currently instructs contributors to fork, branch, and "Create a Github Pull Request from your fork branch to the **master** branch of this repository." This contradicts the release/hotfix branching model above (confirmed against real branches in this repo, e.g. `origin/rc/3.4.0`, `origin/hotfix/3.1.1`) and should be corrected to target the relevant `rc/`/`hotfix/` branch instead — flagging here rather than editing since that file wasn't otherwise in scope for this change.
```
SergeySalinAtDataart: I think this note about inconsistency can be removed but we need to clarify the branching - we do target master directly from our feature branches

CLAUDE.md
```diff
@@ -0,0 +1,102 @@
...
+
+Test projects follow a strict per-component split, each with different infra requirements:
+- `*.Test` — pure unit tests, no external dependencies.
+- `*.Test.Integration` — hits a real SQL database; Azure Storage Emulator (Azurite) must be running; other dependencies (HMD, etc.) are stubbed.
+- `*.Test.Validation` (Matching, Match Prediction) — spins up a real in-memory server and exercises it over HTTP; written in Gherkin via Reqnroll for non-technical review by the Search/BioInformatics teams; requires user secrets and a specific HLA Metadata Dictionary version to be generated locally first.
+- `*.Test.Performance` / `*.Test.Verification` — manual/CI-nightly performance harnesses, gated by the custom `IgnoreExceptOnCiPerfTestAttribute` (only runs when `RUN_CI_PERF_TESTS` env var is set).
+
+There is no local Azure Service Bus emulator — local dev requires a real Azure Service Bus namespace with manually-created topics/subscriptions (see `README_DevelopmentStartUpGuide.md` for the exact topic/subscription names needed per workflow).
```
SergeySalinAtDataart: There's no local Azure Service Bus emulator in the project, however, Microsoft created a real service bus emulator provided in a docker container, supporting up to 10 simultaneous requests and very clunky to configure which we can use but currently don't. I'm not sure we need to mention it here

CLAUDE.md
```diff
@@ -0,0 +1,102 @@
...
+
+### Settings
+
+Non-Functions projects use `appsettings.json` + user secrets. Functions apps use `local.settings.json`, which is scaffolded from a checked-in `local.settings.template.json` on build (never edit `local.settings.json` in a way you expect to be preserved by git — it's gitignored; add new settings to the `.template.json` instead).
+
+## Conventions
+
+- Test framework: NUnit (`[TestFixture]`/`[Test]`), NSubstitute for mocking, AwesomeAssertions (FluentAssertions-compatible, `.Should()`) for assertions, AutoFixture (via a repo `FixtureBuilder.For<T>()` helper) plus custom `*Builder` classes under `TestHelpers/Builders` for test data. Test naming: `MethodUnderTest_ExpectedBehaviour`.
+- Commit messages are prefixed with the GitHub issue number and a semantic type, e.g. `feature: #34: add match prediction`, `fix: #45: fix null reference` (see `README_Contribution_Versioning.md` for the full type list).
```
SergeySalinAtDataart: Now that we've already moved from GitHub issues into Jira we no longer use issue numbers and use ATL-### convention for naming. Should we mention that?

SergeySalinAtDataart (approved):

CLAUDE.md
```diff
@@ -0,0 +1,102 @@
...
+
+## Build and test commands
+
+```bash
+# Restore + build (mirrors build-pipeline.yml)
+dotnet restore
+dotnet build --configuration Release --no-restore
+
+# Run all unit tests (Test projects only, excludes Integration/Validation/Performance/Verification)
+dotnet test "**/*Test.csproj"
```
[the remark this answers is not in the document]
da-ikovalenko: Fixed in 9c2f119c0 — added `shopt -s globstar` and unquoted the glob so the shell actually expands it before `dotnet test` sees it.

CLAUDE.md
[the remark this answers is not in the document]
da-ikovalenko: Fixed spelling — this whole callout is also removed in 9c2f119c0 since it turned out not to be an inconsistency (see other thread on this line).

da-ikovalenko: Good catch — you're right, and this matches what I confirmed separately: master is the actual target for now, with the rc/hotfix flow taking effect once release 4.0 is cut. Rewrote the Branching section in 9c2f119c0 to say that explicitly and removed the inconsistency callout, since README_Contribution_Versioning.md's master-targeting instruction is correct as-is.

CLAUDE.md
da-ikovalenko: Fair point, thanks — added a clarifying clause in 9c2f119c0 noting Microsoft's dockerized emulator exists but isn't adopted here (10 concurrent request cap, clunky config), so it's clear the line is about local practice, not that no such tool exists.

CLAUDE.md
da-ikovalenko: Yes, good call — updated in 9c2f119c0 to describe the current `ATL-###` Jira convention (based on real recent commit history) and flagged that README_Contribution_Versioning.md still documents the older GitHub-issue-number convention.

daria-sorokina-da (approved):

PR #1508, opened 2026-08-20
feature: ATL-282: Extract Data Refresh into dedicated Function App

OTarasov-DataArt:
This change is [Reviewable](https://reviewable.io/reviews/Anthony-Nolan/Atlas/1508)


Atlas.MatchingAlgorithm.Functions/host.json
```diff
@@ -1,10 +1,4 @@
 {
-  "extensions": {
-    "serviceBus": {
-      "maxAutoLockRenewalDuration": "2.00:30:00"
-    }
-  },
-  "functionTimeout": "2.00:00:00",
   "logging": {
     "applicationInsights": {
       "samplingSettings": {
```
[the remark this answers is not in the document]
OTarasov-DataArt: Valid, RunSearch is still a Service Bus trigger here, so bare defaults (5-min lock renewal) would kill/duplicate slow searches. Set functionTimeout: 02:00:00 / maxAutoLockRenewalDuration: 02:30:00 instead. Provisional, we need to discuss the exact value (AC #3's "revert to defaults" probably needs amending).

Atlas.MatchingAlgorithm.Functions.DataRefresh/Program.cs
```diff
@@ -0,0 +1,52 @@
...
+var host = new HostBuilder()
+    .ConfigureFunctionsWebApplication(builder =>
+    {
+        builder.Services.Configure<WorkerOptions>(options =>
+        {
+            var settings = NewtonsoftJsonObjectSerializer.CreateJsonSerializerSettings();
+            settings.ContractResolver = new CamelCasePropertyNamesContractResolver();
+            settings.NullValueHandling = NullValueHandling.Ignore;
+
+            options.Serializer = new NewtonsoftJsonObjectSerializer();
```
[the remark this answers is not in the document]
OTarasov-DataArt: Correct, but pre-existing, copied verbatim from the main app's Program.cs. Out of scope for this extraction (would change SB deserialization behaviour)

da-ikovalenko:
Minor nit: `Atlas.MatchingAlgorithm.Functions.DataRefresh.csproj` still has the `<None Update="ApplicationInsights.config">` entry copied from the `DonorManagement` precedent, but the actual `ApplicationInsights.config` placeholder file (`<!-- Empty file to allow Visual Studio to collect AI telemetry when debugging locally -->`) was never added to this new project — it exists in both the main app and `DonorManagement`. Harmless for the build (an `Update` on a non-matching file is a no-op), but AI telemetry-on-debug won't work locally for this app until it's added. Easy one-line fix: add the same placeholder file to `Atlas.MatchingAlgorithm.Functions.DataRefresh/`.

da-ikovalenko (approved):

OTarasov-DataArt:
Good catch, thanks. I added `ApplicationInsights.config` to `Atlas.MatchingAlgorithm.Functions.DataRefresh`

Atlas.MatchingAlgorithm.Functions.DataRefresh/local.settings.template.json
```diff
@@ -0,0 +1,56 @@
+{
+  "IsEncrypted": false,
+  "Values": {
+    "ApplicationInsights:LogLevel": "Info",
```
SergeySalinAtDataart, line 4: Are you sure we need all of the settings in here for Data Refresh? What about settings in the old file?

Atlas.MatchingAlgorithm.Functions/host.json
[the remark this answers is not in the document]
SergeySalinAtDataart: Let's leave the current values (2 hours and 2.5 hours respectively) and do the tests for Search and HMD regeneration anyway after deployments. If anything fails, we will update the values and redeploy the applications. So I'd vote for resolving this one

Atlas.MatchingAlgorithm.Functions/Startup.cs
```diff
@@ -59,11 +44,8 @@ private static void RegisterSettings(IServiceCollection services)
         {
             services.RegisterAsOptions<ApplicationInsightsSettings>("ApplicationInsights");
             services.RegisterAsOptions<AzureAuthenticationSettings>("AzureManagement:Authentication");
-            services.RegisterAsOptions<AzureDatabaseManagementSettings>("AzureManagement:Database");
             services.RegisterAsOptions<AzureMonitoringSettings>("AzureManagement:Monitoring");
```
SergeySalinAtDataart, line 47: Just for good measure, can you double-check if all the settings and services left in the DI are necessary and all the necessary settings and services are injected in .DataRefresh functions?

Atlas.MatchingAlgorithm.Functions.DataRefresh/ApplicationInsights.config
```diff
@@ -0,0 +1 @@
+﻿<!-- Empty file to allow Visual Studio to collect AI telemetry when debugging locally -->
```
SergeySalinAtDataart: I think we don't really need this file. It feels like a residue and I can't see a better time to remove it than now

SergeySalinAtDataart (changes requested):

Atlas.MatchingAlgorithm.Functions/Startup.cs
OTarasov-DataArt: Checked it. All 10 registered settings map 1:1 to what RegisterDataRefresh consumes, and I verified the whole container resolves by running the app under func start (the isolated host does validate-on-build, so it constructs every registered service).

That run actually surfaced a real problem: ActiveRepositoryFactory (pulled in via RegisterCommonServices) depends on IMatchingAlgorithmSearchLogger, which was only ever registered by the search path. In the combined app that path was always present. Standalone, DI validation failed and the host wouldn't start. Fixed by registering the search logger in RegisterDataRefreshServices

Atlas.MatchingAlgorithm.Functions.DataRefresh/ApplicationInsights.config
OTarasov-DataArt: Agreed, removed it. (The main and DonorManagement apps still have the same file if we want to clean those up later.)

Atlas.MatchingAlgorithm.Functions.DataRefresh/local.settings.template.json
OTarasov-DataArt: Yep, these are the same settings the data refresh path already used, just moved over. Good catch on the old file though, it still had the now-unused DataRefresh: and AzureManagement:Database: keys, so I cleaned those out too.

da-ikovalenko:
Real search-duration data from `SearchTracking.SearchRequestMatchingAlgorithmAttempts` shows `TotalDurationMinutes` for successful search attempts reaching as high as **596 minutes (~9h56m)** — well beyond the 2-hour `functionTimeout` (`02:00:00`) / 2.5-hour `maxAutoLockRenewalDuration` (`02:30:00`) this PR sets on the main app's `host.json`.

Since `SearchFunctions.RunSearch` is the only Service-Bus-triggered function remaining in `Atlas.MatchingAlgorithm.Functions` after this extraction, these values need to be sized against the observed worst case rather than an assumed 2h ceiling. Suggest bumping to something like `functionTimeout: "12:00:00"` / `maxAutoLockRenewalDuration: "12:30:00"` (keeping the same 30-minute renewal buffer already used in `Atlas.MatchingAlgorithm.Functions.DataRefresh/host.json`) so there's real headroom above the ~10h max observed today.

Updating ATL-282 AC #3 to reflect this.

Atlas.MatchingAlgorithm.Functions/host.json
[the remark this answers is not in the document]
daria-sorokina-da: Here's the data from LIVE WMDA:



Some of the matching attempts take a long time

daria-sorokina-da (changes requested):
> @daria-sorokina-da partially reviewed 14 files and all commit messages, and made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/Anthony-Nolan/Atlas/1508)* status: all files reviewed, 5 unresolved discussions (waiting on OTarasov-DataArt and SergeySalinAtDataart).

___
*[`Atlas.MatchingAlgorithm.Functions.DataRefresh/Atlas.MatchingAlgorithm.Functions.DataRefresh.csproj` line 1 at r4](https://reviewable.io/reviews/Anthony-Nolan/Atlas/1508#-P-s_LT4DYJjD5RXyEPB:-P-s_LT4DYJjD5RXyEPC:b5m47a0) ([raw file](https://github.com/Anthony-Nolan/Atlas/blob/5c438378426c86d780de3cef6535e96238f8101d/Atlas.MatchingAlgorithm.Functions.DataRefresh/Atlas.MatchingAlgorithm.Functions.DataRefresh.csproj#L1)):*
> ```XML
> <Project Sdk="Microsoft.NET.Sdk">
> ```

Is there a separate task to add a new app to terraform?

OTarasov-DataArt:
*[`Atlas.MatchingAlgorithm.Functions.DataRefresh/Atlas.MatchingAlgorithm.Functions.DataRefresh.csproj` line 1 at r4](https://reviewable.io/reviews/Anthony-Nolan/Atlas/1508#-P-s_LT4DYJjD5RXyEPB:-P-tCdChArPizvNsS_e4:b4qb0nw) ([raw file](https://github.com/Anthony-Nolan/Atlas/blob/5c438378426c86d780de3cef6535e96238f8101d/Atlas.MatchingAlgorithm.Functions.DataRefresh/Atlas.MatchingAlgorithm.Functions.DataRefresh.csproj#L1)):*

Is there a separate task to add a new app to terraform?
</blockquote></details>

Yes, I believe it's [ATL-301](https://anthonynolan.atlassian.net/browse/ATL-301)





[ATL-301]: https://anthonynolan.atlassian.net/browse/ATL-301?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

Atlas.MatchingAlgorithm.Functions/host.json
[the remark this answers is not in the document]
OTarasov-DataArt: Bumped the values to `functionTimeout: "12:00:00"` and `maxAutoLockRenewalDuration: "12:30:00"`

daria-sorokina-da (approved):
> @daria-sorokina-da reviewed 1 file and all commit messages, and resolved 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/Anthony-Nolan/Atlas/1508)* status: all files reviewed, 3 unresolved discussions (waiting on SergeySalinAtDataart).

SergeySalinAtDataart (approved):

PR #1513, opened 2026-08-26
feature: ATL-233: stop the exception path for an expected HLA lookup failure

SergeySalinAtDataart:
The HF sets are encoded at an older nomenclature version than the refresh runs, so a name with no data at that version is the expected case, not an error. It cost a cached-miss repeat, a throw, a second throw wrapping it, and a multi-kilobyte exception string per occurrence. This removes all four.

Three changes, one path:

1. MetadataServiceBase now caches the lookup OUTCOME, not only a hit, so a name with no data is remembered.

2. IHlaMetadataDictionary gains three additive methods - TryConvertHla, TryConvertGGroupToPGroup, TryConvertSmallGGroupToPGroup - over one new MetadataServiceBase.TryGetMetadata, which reads the same cached outcome and returns it as (bool WasFound, T Value). A name with no data now costs one dictionary probe: no InvalidHlaException, no second throw wrapping it, no catch in HlaConverterBase. The throwing methods are untouched, which keeps DonorHlaExpander, SearchRunner and RepeatSearchRunner on the expected-error pathway they use as control flow. ValueOrThrow() and ValueOrNotFound() are two accessors of one cached outcome, so whichever caller meets a bad name first pays for it and the other is served from cache. Only the three categories Match Prediction converts to - G group, P group and small g group - get a non-throwing route; serology and the two-field-allele categories keep the throwing lookup behind a catch, their failures being too rare to earn the surface.

3. The conversion-failure event drops its exception.ToString() dimension - a full async stack, multiple kilobytes - which said nothing the other five dimensions did not. An example 248-donor run reached that path 11,538 times.

BEHAVIOUR CHANGE, deliberate: an infrastructure fault now propagates. GetMetadata used to wrap EVERY exception into HlaMetadataDictionaryException and HlaConverterBase swallowed it, so a failed storage request and a missing name arrived as the same thing - a transient blip read as "this HLA does not exist", and match prediction ran on an incomplete expansion with nothing logged as an error. TryGetMetadata treats only HlaMetadataDictionaryException and its subclasses (InvalidHlaException included) as not-found; anything else escapes, on both routes, with a test for each.

Tests. MetadataServiceBaseTests gains six over a second concrete service (GGroupToPGroupMetadataService, where "in the data, and has no P group" is real): not-found without a throw, the value when found, found-with-a-null-value, one repository call across five not-found lookups, agreement with the throwing path over one cached outcome, and a TimeoutException that propagates and is NOT cached. HMD's HlaConverterTests gains eight for TryConvertHla. GroupToPGroupConverterTests is new, because GGroupToPGroupConverter and SmallGGroupToPGroupConverter had no test at all, and the case that needed one is found-with-no-P-group - a group of null-expressing alleles, which must stay found rather than be logged as a failure and retried at the other version.

- - -
This change is [Reviewable](https://reviewable.io/reviews/Anthony-Nolan/Atlas/1513)

### DIFF
--- a/Atlas.HlaMetadataDictionary.Test/UnitTests/Services/HlaConversion/HlaConverterTests.cs
+++ b/Atlas.HlaMetadataDictionary.Test/UnitTests/Services/HlaConversion/HlaConverterTests.cs
@@ -4,6 +4,7 @@ using System.Threading.Tasks;
 using Atlas.Common.ApplicationInsights;
 using Atlas.Common.GeneticData;
 using Atlas.Common.Public.Models.GeneticData;
+using Atlas.HlaMetadataDictionary.ExternalInterface.Exceptions;
 using Atlas.HlaMetadataDictionary.ExternalInterface.Models;
 using Atlas.HlaMetadataDictionary.ExternalInterface.Models.HLATypings;
 using Atlas.HlaMetadataDictionary.ExternalInterface.Models.Metadata;
@@ -14,6 +15,7 @@ using Atlas.HlaMetadataDictionary.Test.TestHelpers.Builders;
 using Atlas.HlaMetadataDictionary.Test.TestHelpers.Builders.ScoringInfoBuilders;
 using AwesomeAssertions;
 using NSubstitute;
+using NSubstitute.ExceptionExtensions;
 using NUnit.Framework;
 
 namespace Atlas.HlaMetadataDictionary.Test.UnitTests.Services.HlaConversion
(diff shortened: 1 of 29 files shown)


Atlas.HlaMetadataDictionary/Services/DataRetrieval/MetadataServiceBase.cs
```diff
@@ -53,18 +105,93 @@ private static string FormatLookupName(string lookupName)
...
+            public static LookupOutcome NotFound(HlaMetadataDictionaryException exception) =>
+                new(default, ExceptionDispatchInfo.Capture(exception));
+
+            /// <summary>
+            /// The value, or the original exception re-thrown. Captured rather than stored bare so that the first
+            /// failure's stack survives every later re-throw of it, which a plain <c>throw stored;</c> would reset.
+            /// </summary>
+            public T ValueOrThrow()
+            {
+                notFound?.Throw();
```
da-ikovalenko, line 222: **Correctness (possible data race):** A not-found lookup outcome is now cached once and rethrown to every caller via `ExceptionDispatchInfo.Throw()` here. `Throw()` mutates the exception's internal stack-trace state on every call, and this cache sits behind a process-wide singleton `IPersistentCacheProvider`.

**Failure scenario:** Two concurrent callers (e.g. parallel `DonorHlaExpander` batches, or Match Prediction's `Task.WhenAll` over genotypes) look up the same missing HLA name at the same nomenclature version at the same time. Both read the same cached `LookupOutcome` and both call `ValueOrThrow()` -> `notFound.Throw()` on the identical shared exception instance concurrently, racing to mutate its stack-trace string and producing corrupted/interleaved stack traces wherever `e.ToString()` is logged (e.g. `DonorBatchProcessor`). This race did not exist pre-PR, since each caller previously threw a freshly constructed exception.

Atlas.HlaMetadataDictionary/Services/DataRetrieval/MetadataServiceBase.cs
```diff
@@ -53,18 +105,93 @@ private static string FormatLookupName(string lookupName)
...
+            //    entry whose factory threw, and so the next donor carrying the same bad name repeated the whole thing:
+            //    the storage request, two throws, the logged event and - because the HF sets are at nomenclature 3480
+            //    while the refresh runs 3650 - the retry at the other version. 
+            // 2. A lookup that legitimately returns null is now served from the cache. `cache.Get<T>,` cannot tell "not
+            //    cached" from "cached, and the answer is null", so those were re-fetched forever.
+            var cachedOutcome = cache.Get<LookupOutcome>(key);
+            if (cachedOutcome != null)
             {
-                return existingRecord;
+                return cachedOutcome.ValueOrThrow();
```
da-ikovalenko: **Simplification:** `GetOrAddCachedMetadata` and the new `TryGetMetadata` (above, around line 68) duplicate the same cache-key build, lookup, validity check, and `GetOrAddAsync(Lookup)` call, differing only in how the final `LookupOutcome` is unwrapped (`ValueOrThrow()` here vs `ValueOrNotFound()` there).

**Risk:** A future change to the caching strategy (key format, validity check, cache options) made in only one of the two near-identical methods could silently reintroduce a behavioural mismatch between the throwing and non-throwing routes that this PR otherwise keeps in sync via one shared cached outcome.

Atlas.HlaMetadataDictionary.Test/UnitTests/Services/HlaConversion/HlaConverterTests.cs
```diff
@@ -212,6 +214,150 @@ public async Task ConvertHla_TargetIsSerology_ReturnsMatchingSerologies()
...
+            });
+
+            wasFound.Should().BeFalse();
+            hla.Should().BeNull();
+        }
+
+        [TestCase(TargetHlaCategory.GGroup)]
+        [TestCase(TargetHlaCategory.SmallGGroup)]
+        [TestCase(TargetHlaCategory.Serology)]
+        public async Task TryConvertHla_WhenTheLookupFaults_PropagatesInsteadOfReportingNoData(TargetHlaCategory targetHla)
```
da-ikovalenko, line 351: **Test coverage gap:** This fault-propagation test is parametrised only over `GGroup`, `SmallGGroup` and `Serology`, leaving `PGroup` and the two `TwoFieldAllele` categories — which share the same `TryGetHlaScoringInfo`/default-catch code path — untested for the fault-propagation behaviour.

**Risk:** A future change to `TryGetHlaScoringInfo` or the default case's `catch (HlaMetadataDictionaryException)` block that starts swallowing infra faults again for `PGroup` specifically would pass this suite, even though the sibling `GGroup` case (sharing the same helper) would catch it.

Atlas.HlaMetadataDictionary/Services/HlaConversion/HlaConverter.cs
```diff
@@ -13,6 +14,15 @@ namespace Atlas.HlaMetadataDictionary.Services.HlaConversion
     internal interface IHlaConverter
     {
         Task<IReadOnlyCollection<string>> ConvertHla(Locus locus, string hlaName, HlaConversionBehaviour conversionBehaviour);
+
+        /// <summary>
+        /// <see cref="ConvertHla"/> for a caller that treats an unknown HLA name as an answer. See
+        /// <c><seealso cref="MetadataServiceBase{T}.TryGetMetadata"/>MetadataServiceBase.TryGetMetadata</c> for why an infrastructure fault still throws.
```
da-ikovalenko: **Style nit:** This XML doc comment line is 154 characters, exceeding the repo's `.editorconfig` `max_line_length = 150`, and it's redundant: a `<seealso cref="MetadataServiceBase{T}.TryGetMetadata"/>` is immediately followed by the literal text "MetadataServiceBase.TryGetMetadata".

da-ikovalenko (commented):
Automated review findings (via Claude Code).

da-ikovalenko (changes requested):

Atlas.HlaMetadataDictionary/Services/DataRetrieval/MetadataServiceBase.cs
SergeySalinAtDataart: Long story short: Core CLR makes a copy of exception stuff for each thread trying to access it so there is little chance of causing real corruption in stack traces with this approach

Atlas.HlaMetadataDictionary/Services/HlaConversion/HlaConverter.cs
```diff
@@ -66,6 +76,62 @@ public async Task<IReadOnlyCollection<string>> ConvertHla(Locus locus, string hl
...
+                    return (wasFound, wasFound ? smallGGroups.ToList() : null);
+                }
+                default:
+                    try
+                    {
+                        return (true, await ConvertHla(locus, hlaName, conversionBehaviour));
+                    }
+                    catch (HlaMetadataDictionaryException)
+                    {
+                        return (false, null);
```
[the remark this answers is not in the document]
SergeySalinAtDataart: Rewritten the root of the problem - MetadataServiceBase no longer wraps the real infrastructure exceptions into HlaMetadataDictionaryException and doesn't try to hide them - they should propagate now

Atlas.HlaMetadataDictionary/Services/DataRetrieval/MetadataServiceBase.cs
```diff
@@ -53,18 +105,93 @@ private static string FormatLookupName(string lookupName)
...
+        /// </remarks>
+        private async Task<LookupOutcome> Lookup(Locus locus, string lookupName, string hlaNomenclatureVersion)
+        {
+            try
+            {
+                return LookupOutcome.Found(await PerformLookup(locus, lookupName, hlaNomenclatureVersion));
+            }
+            catch (HlaMetadataDictionaryException nameNotFound)
+            {
+                return LookupOutcome.NotFound(nameNotFound);
```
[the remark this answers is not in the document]
SergeySalinAtDataart: Rewritten the root of the problem - MetadataServiceBase no longer wraps the real infrastructure exceptions into HlaMetadataDictionaryException and doesn't try to hide them - they should propagate now

Atlas.HlaMetadataDictionary/Services/DataRetrieval/MetadataServiceBase.cs
SergeySalinAtDataart: Updated to reuse the same code

Atlas.HlaMetadataDictionary/Services/HlaConversion/HlaConverter.cs
SergeySalinAtDataart: Updated

Atlas.HlaMetadataDictionary.Test/UnitTests/Services/HlaConversion/HlaConverterTests.cs
SergeySalinAtDataart: Added a PGroup test case and a separate test for TwoFieldAllele

daria-sorokina-da (approved):
> @daria-sorokina-da reviewed 29 files and all commit messages, and resolved 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/Anthony-Nolan/Atlas/1513)* status: all files reviewed, 4 unresolved discussions (waiting on da-ikovalenko and OTarasov-DataArt).

da-ikovalenko (approved):

PR #1515, opened 2026-08-27
feature: ATL-233: genotype imputation expansion

SergeySalinAtDataart:
A set of improvements targeting genotype imputation performance

- - -
This change is [Reviewable](https://reviewable.io/reviews/Anthony-Nolan/Atlas/1515)


daria-sorokina-da (changes requested):
> @daria-sorokina-da reviewed 8 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/Anthony-Nolan/Atlas/1515)* status: 8 of 20 files reviewed, 3 unresolved discussions (waiting on da-ikovalenko, OTarasov-DataArt, and SergeySalinAtDataart).

___
*[`Atlas.MatchPrediction/Services/MatchProbability/GenotypeImputationService.cs` line 39 at r1](https://reviewable.io/reviews/Anthony-Nolan/Atlas/1515#-P03A1nm97Z2ILfs3u4r:-P03A1nm97Z2ILfs3u4s:b-kxuplo) ([raw file](https://github.com/Anthony-Nolan/Atlas/blob/c7c60a85d25d1e0eb6a04a4d422b0cf011dbf013/Atlas.MatchPrediction/Services/MatchProbability/GenotypeImputationService.cs#L39)):*
> ```Smalltalk
>         private readonly GenotypeImputationSettings settings;
> 
>         // ATL-233 T2 moved likelihood calculation into the expansion, where each genotype's haplotype pair is still in
> ```

@da-ikovalenko @SergeySalinAtDataart 

Are such comments helpful here, in this specific place in the code? 

In this particular case. we removed the diplotype calculator (it's not in the code anymore), so this will not give me any information about the current state of things.

Should we instead hold such comments in an ADR, or maybe we lack some new audit of algorithm changes?

Tbh, as a reviewer, it is very hard for me to follow the train of thought through the comments scattered across different files - one "audit" record per ticket in the PR would be more helpful, and the comments could then be shorter and more precise - not about "changes made", but any additional information about the piece of code.

What do you think?

Atlas.MatchPrediction/Services/CompressedPhenotypeExpansion/CompressedPhenotypeExpander.cs
```diff
@@ -92,125 +242,408 @@ private static bool IsUnambiguousAtAllowedLoci(
...
+    /// </para>
+    ///
+    /// <para>
+    /// It resolves for every survivor, including any that no kept pair ends up using - so a donor whose typing is not
+    /// representable at all now pays S lookups where it previously paid none. Accepted rather than overlooked: filling
+    /// the array lazily would put a branch on the O(S²) pairing path, which is a worse trade than S lookups, and those
+    /// donors are 1.28% of the corpus with a near-zero <c>Pair</c> share (§4), i.e. a small S.
+    /// </para>
+    /// </summary>
+    private async Task<decimal[]> ResolveHaplotypeFrequencies(
```
da-ikovalenko, line 432: `ResolveHaplotypeFrequencies` awaits `GetFrequencyForHla` sequentially in a `for` loop instead of overlapping the independent per-survivor lookups. For a donor with a large survivor count (this PR's own numbers cite up to 6,037), each lookup is an independent cache-backed async call - a smaller-scale version of exactly the "awaited lookup fan-out" cost this PR spends most of its effort eliminating elsewhere in the same file. Worth a bounded `Task.WhenAll`/batching pass, non-blocking.

Atlas.MatchPrediction/Services/CompressedPhenotypeExpansion/CompressedPhenotypeExpander.cs
```diff
@@ -92,125 +242,408 @@ private static bool IsUnambiguousAtAllowedLoci(
...
+    /// without materialising a <c>List&lt;Locus&gt;</c>, a <c>Diplotype</c> or the string phenotype.
+    /// </para>
+    ///
+    /// <para>
+    /// T3's strong form takes the two haplotypes in their name form, which is what this always compared: the
+    /// <c>?.Hla</c> the category form needed was the only reason it took that form at all. A stored haplotype may be
+    /// untyped at an allowed locus, so a name may be null, and two nulls compare equal exactly as before.
+    /// </para>
+    /// </summary>
+    private static int HomozygosityCorrectionFactor(
```
da-ikovalenko, line 471: `HomozygosityCorrectionFactor` re-implements the same "which allowed loci differ between two haplotypes" logic that already exists, independently, as `UnambiguousGenotypeExpander.GetHeterozygousLoci`. Two independent implementations of a clinically-relevant heterozygosity test (one on a pair of `LociInfo<string>` returning 1/2, the other on a single `PhenotypeInfo<string>` returning `List<Locus>`) must be kept semantically identical by hand - a future change to one (new locus, different null handling) can silently diverge from the other. Consider consolidating, non-blocking.

Atlas.MatchPrediction/Services/HaplotypeFrequencies/FrequencySetCacheEntry.cs
```diff
@@ -25,4 +30,81 @@ public FrozenDictionary<HaplotypeKey, decimal> ConsolidatedFrequencies
         get => consolidatedFrequencies;
         set => consolidatedFrequencies = value;
     }
+
+    // volatile, and assigned with ??= rather than under a lock: the read-test-write is not atomic, so two threads
+    // racing the first access can both project - but a FrozenDictionary enumerates in a stable order, so the two
+    // results are element-for-element identical and either may win. A reference write has release semantics, so the
+    // collection is fully built before any thread can observe it.
+    private volatile DataByResolution<HaplotypeKey[]> projectedPool;
```
da-ikovalenko, line 40: The `volatile` + unlocked `??=` lazy-init here defends against a race that the only production caller already forecloses: `HaplotypeFrequencyCache.BuildEntryFromDatabase` builds this eagerly (`_ = entry.ProjectedPool;`) inside `cache.GetOrAddAsync`, before the entry ever escapes. The safety property ("racing builds agree because `FrozenDictionary` enumerates in stable order") is real but lives one layer up from where the comment claims it does - if `GetOrAddAsync`'s single-build guarantee is ever weakened, or the eager call removed, a maintainer reading only this file would wrongly trust the `volatile`/`??=` pattern as sufficient. Non-blocking, but worth tightening the comment or moving the guarantee where it's actually enforced.

Atlas.MatchPrediction/Services/CompressedPhenotypeExpansion/CompressedPhenotypeExpander.cs
```diff
@@ -92,125 +242,408 @@ private static bool IsUnambiguousAtAllowedLoci(
...
-            foreach (var locus in allowedLociArray)
-            {
-                var hla1 = haplotype1.GetLocus(locus);
-                var hla2 = haplotype2.GetLocus(locus);
+            haplotypeNames[h] = haplotypeList[h].Map(hla => hla?.Hla);
+        }
+
+        // ATL-233 T3: a List, not a HashSet - see the ExpandedGenotypes remarks for why this can never de-duplicate.
+        // genotypeHlaNames is kept index-aligned with it so truncation needs no second ToHlaNames() per genotype.
+        var genotypePairs = new List<GenotypePair>();
```
da-ikovalenko, line 297: `genotypePairs`, `genotypeHlaNames` and `likelihoods` are never pre-sized before the O(S²) pairing loop that fills them, despite `haplotypeList.Count` being known beforehand. For large survivor counts these grow via repeated internal array doubling (both `List<T>` and `Dictionary<K,V>` reallocate+copy on growth) across up to millions of entries corpus-wide - a capacity hint would remove most of that reallocation overhead in exactly the hot path this PR optimizes. Non-blocking.

da-ikovalenko (commented):
Automated review pass (high effort). No correctness bugs found in the imputation-equivalence-critical paths (bitmask pairing, id-indexed pool filter, PriorityQueue truncation tie-break, cached `ProjectedPool` ordering) - all verified by hand against the shipped `ImputationEquivalenceTests.cs` oracle suite. Four non-blocking cleanup items below.

**Also flagging a ticket-linkage issue, not code:** this PR's branch/title cite ATL-233, but that Jira ticket is "Precompute on Data Refresh - new stage 55: dedup & dispatch to ACA" (still in Backlog, redesigned 2026-08-26) - unrelated to this change. The actual content here (uncached per-donor pool rebuild fix, bounded truncation replacing `OrderByDescending().Take()`) matches ATL-303 and ATL-304, the Match Prediction OOM root-cause bugs. Recommend retitling/relinking to ATL-303/ATL-304 so those tickets actually close.

da-ikovalenko (changes requested):

Atlas.MatchPrediction/Services/CompressedPhenotypeExpansion/CompressedPhenotypeExpander.cs
SergeySalinAtDataart: There's a problem that we don't know exactly how many survivors we need to keep and we usually only keep a small fraction of pairs after the loop, so the capacity hint will not help us here - we will either undershoot or overshoot it. We can do another O(S^2) to calculate exact sizes by doing cheap integer arithmetic if that's what you wish

Atlas.MatchPrediction/Services/CompressedPhenotypeExpansion/CompressedPhenotypeExpander.cs
SergeySalinAtDataart: Unfortunately, because of the cache nature we infect the call with async despite only the first call being truly asynchronous - all consequent calls return Task.FromResult and we pay only for async/await machinery without actually awaiting anything. Task.WhenAll or batching won't help us here.

Atlas.MatchPrediction/Services/CompressedPhenotypeExpansion/CompressedPhenotypeExpander.cs
SergeySalinAtDataart: I would argue those methods answer different questions, consumed by different endpoints (production versus debug) and covered by different test suites. Consolidating them is a completely different task from performance improvements for genotyp imputation and while I, personally, love refactoring, I cannot justify putting that here

Atlas.MatchPrediction/Services/CompressedPhenotypeExpansion/CompressedPhenotypeExpander.cs
```diff
@@ -38,9 +39,82 @@ internal interface ICompressedPhenotypeExpander
...
+/// frequency is a pure function of (set, haplotype, excluded loci), needing one resolution per <i>survivor</i>. The
+/// shipped code instead resolved two per <i>genotype</i>, downstream, after the pairing had thrown the pair away:
+/// 136,879 awaited lookups for a tail donor against 465.7 survivors, each re-entering LazyCache's
+/// <c>GetOrAddAsync</c>. That machinery, not the arithmetic, was 39 of the 40.9 s the Likelihood phase measured.
+/// </para>
+///
+/// <para>
+/// <see cref="Likelihoods"/> is keyed by the genotype's HLA <i>names</i>, matching
+/// <c>ImputedGenotypes.GenotypeLikelihoods</c>, so its count is the DISTINCT genotype count while
+/// <see cref="Genotypes"/> keeps typing category and so may hold more. Keying it here also spares one
```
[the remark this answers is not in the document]
SergeySalinAtDataart: Comments updated

Atlas.MatchPrediction/Services/CompressedPhenotypeExpansion/CompressedPhenotypeExpander.cs
```diff
@@ -92,125 +242,408 @@ private static bool IsUnambiguousAtAllowedLoci(
...
         var allowedLociArray = allowedLoci.ToArray();
 
+        // ATL-233 T2. This selects WHICH frequency a haplotype has - with loci excluded, a haplotype stands for a
+        // group of stored haplotypes whose frequencies are consolidated - and it is constant for the whole request.
+        // The pre-T2 DiplotypeLikelihoodCalculator (since deleted) rebuilt it per genotype, twice over, because LocusSettings.MatchPredictionLoci
+        // is an expression-bodied property that re-enumerates the enum and allocates a fresh HashSet on every read.
+        var excludedLoci = LocusSettings.MatchPredictionLoci.Except(allowedLoci).ToHashSet();
+
+        // One resolution per survivor, instead of two per genotype further downstream.
+        var frequencies = await ResolveHaplotypeFrequencies(hfSetId, haplotypeList, excludedLoci);
```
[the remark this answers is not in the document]
SergeySalinAtDataart: There will be a further improvement on top of this branch that already works with awaiting instead of fire-and-forgetting lookup warm-up

daria-sorokina-da (approved):
> @daria-sorokina-da reviewed 19 files and all commit messages, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/Anthony-Nolan/Atlas/1515)* status: all files reviewed, 6 unresolved discussions (waiting on da-ikovalenko, OTarasov-DataArt, and SergeySalinAtDataart).

Atlas.MatchPrediction/Services/HaplotypeFrequencies/FrequencySetCacheEntry.cs
SergeySalinAtDataart: Comment tightened!

da-ikovalenko (approved):

PR #1516, opened 2026-08-27
Atl 258 data refresh run level lease

OTarasov-DataArt:
This change is [Reviewable](https://reviewable.io/reviews/Anthony-Nolan/Atlas/1516)


Atlas.MatchingAlgorithm/Services/DataRefresh/DonorImport/DonorImporter.cs
```diff
@@ -66,14 +71,15 @@ public DonorImporter(
             this.donorReader = donorReader;
         }
 
-        public async Task ImportDonors(bool shouldMarkDonorsAsUpdated)
+        public async Task ImportDonors(bool shouldMarkDonorsAsUpdated, CancellationToken cancellationToken)
```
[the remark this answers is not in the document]
OTarasov-DataArt: Defaults are on the IDonorImporter interface, as they were before this change. Callers use the interface, so nothing breaks, test project builds clean.

Atlas.MatchingAlgorithm.Data.Persistent/Migrations/20260824135502_AddLeaseColumnsToDataRefreshHistory.cs
```diff
@@ -0,0 +1,43 @@
...
+
+namespace Atlas.MatchingAlgorithm.Data.Persistent.Migrations
+{
+    /// <inheritdoc />
+    public partial class AddLeaseColumnsToDataRefreshHistory : Migration
+    {
+        /// <inheritdoc />
+        protected override void Up(MigrationBuilder migrationBuilder)
+        {
+            migrationBuilder.AddColumn<DateTime>(
```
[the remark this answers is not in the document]
OTarasov-DataArt: Declining, generated by dotnet ef migrations add. nullable: true already makes the column nullable and the SQL is identical. No migration in this repo uses a nullable generic.

Atlas.MatchingAlgorithm.Data.Persistent/Repositories/DataRefreshHistoryRepository.cs
```diff
@@ -100,6 +125,47 @@ public async Task MarkStageAsComplete(DataRefreshRecord record, DataRefreshStage
...
+                            && (r.LeaseOwner == null || r.LeaseOwner == owner || r.LeaseExpiresUtc < nowUtc))
+                .ExecuteUpdateAsync(s => s
+                    .SetProperty(r => r.LeaseOwner, owner)
+                    .SetProperty(r => r.LeaseExpiresUtc, expiry));
+
+            return rowsUpdated == 1;
+        }
+
+        /// <inheritdoc />
+        public async Task<bool> TryRenewRefreshLease(int recordId, Guid owner, DateTime nowUtc, TimeSpan ttl)
```
SergeySalinAtDataart: I'm not quite sure why do we pass nowUtc here with additional time span only to combine them into expiry? I believe we can do that in the caller method

Atlas.MatchingAlgorithm.Test.Integration/IntegrationTests/DataRefresh/DataRefreshLeaseTests.cs
```diff
@@ -0,0 +1,228 @@
...
+            var record = await ReadRecord(recordId);
+            record.LeaseOwner.Should().Be(owner);
+            record.LeaseExpiresUtc.Should().BeCloseTo(now + LeaseDuration, TimeSpan.FromSeconds(1));
+        }
+
+        [Test]
+        public async Task TryClaimRefreshLease_WhenAnotherInvocationHoldsALiveLease_DoesNotClaimIt()
+        {
+            var recordId = await AnOpenRefreshRecord();
+            var incumbent = Guid.NewGuid();
```
SergeySalinAtDataart: I think we can call it "owner" or "current holder"

Atlas.MatchingAlgorithm/Services/DataRefresh/DataRefreshOrchestrator.cs
```diff
@@ -63,18 +76,211 @@ public DataRefreshOrchestrator(
...
             this.dataRefreshCompletionNotifier = dataRefreshCompletionNotifier;
+            this.serviceScopeFactory = serviceScopeFactory;
+        }
+
+        public async Task OrchestrateDataRefresh(int dataRefreshRecordId, Guid? invocationId)
+        {
+            var (leaseDuration, renewalInterval) = ValidatedLeaseTimings();
+            var leaseOwner = invocationId ?? Guid.NewGuid();
+
+            if (!await dataRefreshHistoryRepository.TryClaimRefreshLease(dataRefreshRecordId, leaseOwner, DateTime.UtcNow, leaseDuration))
```
SergeySalinAtDataart, line 87: What happens if our run dies and then we try to resubmit/redeliver the message again while DataRefresh record is still leased? Will we just get an Application Insights message and that's it?

Atlas.MatchingAlgorithm/Services/DataRefresh/DataRefreshOrchestrator.cs
```diff
@@ -63,18 +76,211 @@ public DataRefreshOrchestrator(
...
+            }
+
+            return (leaseDuration, renewalInterval);
+        }
+
+        private async Task LogRefusedClaim(int dataRefreshRecordId, Guid leaseOwner)
+        {
+            // Deliberately does not throw: throwing here would propagate out of the function and cause the very Service
+            // Bus redelivery that this lease exists to make harmless.
+            var incumbent = await TryDescribeIncumbentLease(dataRefreshRecordId);
```
SergeySalinAtDataart: I still think that "owner" is better than "incumbent" 😅

Atlas.MatchingAlgorithm/Services/DataRefresh/DataRefreshOrchestrator.cs
```diff
@@ -63,18 +76,211 @@ public DataRefreshOrchestrator(
...
+                    logger.SendTrace(
+                        $"{LoggingPrefix} Invocation {leaseOwner} no longer holds the lease on record {recordId}; it has been taken over by " +
+                        "another invocation. Aborting this run.", LogLevel.Critical);
+                    await cancellationTokenSource.CancelAsync();
+                    return;
+                }
+                catch (Exception e)
+                {
+                    // Being unable to reach the database is not the same as being fenced, and must not be treated as
+                    // such. At the default timings the lease spans 30 renewal attempts, so 29 consecutive failures are
```
SergeySalinAtDataart: I'm not sure but maybe we can implement some kind of exponential backoff here? 30 minutes should allow us for quite a several attempts of reaching a database for lease renewal, but I think EF Core should already have something in place, like execution strategy or smth? Can you check if it's a good fit here please?

SergeySalinAtDataart (changes requested):

Atlas.MatchingAlgorithm/Services/DataRefresh/DonorImport/DonorImporter.cs
```diff
@@ -66,14 +71,15 @@ public DonorImporter(
...
         {
             try
             {
                 var allFailedDonors = new List<FailedDonorInfo>();
                 var donorsStream = donorReader.StreamAllDonors().Select(d => d.MapImportDonorToMatchingUpdateDonor());
                 foreach (var streamedDonorBatch in donorsStream.Batch(BatchSize))
                 {
+                    cancellationToken.ThrowIfCancellationRequested();
```
da-ikovalenko, line 82: No unit test exercises this check directly — the only new tests are in `DataRefreshOrchestratorTests`, which mocks `IDataRefreshRunner.RefreshData` entirely, so this line never actually runs under test. The ticket calls the cancellation threading out as "the largest part of this ticket's diff" and asks reviewers to be explicit about it — worth a test asserting a cancelled token stops the import at a batch boundary (and that it propagates as `OperationCanceledException`, not wrapped in `DonorImportHttpException`).

Atlas.MatchingAlgorithm/Services/DataRefresh/HlaProcessing/HlaProcessor.cs
```diff
@@ -160,6 +174,8 @@ private async Task PerformHlaUpdate(
 
                 await foreach (var donorBatch in batchedDonors)
                 {
+                    cancellationToken.ThrowIfCancellationRequested();
```
da-ikovalenko, line 177: Same gap as DonorImporter: nothing under test drives this check, since `DataRefreshOrchestratorTests` mocks `RefreshData` before it ever reaches here. Given the ticket explicitly flags cancellation threading as the largest and most review-worthy part of this diff, a test that cancels mid-`await foreach` and asserts the run stops on a batch boundary (without corrupting `LastSafelyProcessedDonor`) would close the loop.

Atlas.MatchingAlgorithm.Data.Persistent/Migrations/SearchAlgorithmPersistentContextModelSnapshot.cs
```diff
@@ -18,18 +18,18 @@ protected override void BuildModel(ModelBuilder modelBuilder)
 #pragma warning disable 612, 618
             modelBuilder
                 .HasDefaultSchema("MatchingAlgorithmPersistent")
-                .HasAnnotation("ProductVersion", "6.0.3")
+                .HasAnnotation("ProductVersion", "10.0.9")
```
da-ikovalenko, line 21: This bumps `ProductVersion` from 6.0.3 to 10.0.9 and drops the explicit `(1L, 1)` identity-seed args across *every* entity in the snapshot, not just the new lease columns — looks like it was regenerated with a different local `dotnet-ef` tool version than whatever last touched this file. Functionally inert (defaults are seed=1/increment=1 either way), but flagging so it doesn't cause confusing diffs or a spurious conflict for the next migration someone else generates.

Atlas.MatchingAlgorithm/Services/DataRefresh/DataRefreshOrchestrator.cs
OTarasov-DataArt: Checked, EF Core already has it: EnableRetryOnFailure(5, TimeSpan.FromSeconds(5), null) in ContextFactory.cs:44, so transient failures are already retried with backoff inside each renewal call. What reaches this catch is either non-transient or has already exhausted those retries.
I'm not sure if we should add more on top though: the fixed 60s tick is what gives 30 attempts per 30-minute lease. Backing off exponentially would spend the same lease on about 6 attempts, making it potentially more likely to be lost, not less

Atlas.MatchingAlgorithm.Data.Persistent/Repositories/DataRefreshHistoryRepository.cs
OTarasov-DataArt: Done for TryRenewRefreshLease, it takes expiry now. TryClaimRefreshLease keeps both, since it also uses nowUtc on its own to check if the existing lease has expired

Atlas.MatchingAlgorithm.Test.Integration/IntegrationTests/DataRefresh/DataRefreshLeaseTests.cs
OTarasov-DataArt: Renamed to "owner"

Atlas.MatchingAlgorithm/Services/DataRefresh/DataRefreshOrchestrator.cs
OTarasov-DataArt: Renamed

SergeySalinAtDataart (approved):

Atlas.MatchingAlgorithm/Services/DataRefresh/DonorImport/DonorImporter.cs
OTarasov-DataArt: Added three tests: cancels during the first of two batches and asserts one insert, asserts cancellation propagates as OperationCanceledException rather than wrapped as DonorImportHttpException, and a cancelled-before-start case.

Atlas.MatchingAlgorithm/Services/DataRefresh/HlaProcessing/HlaProcessor.cs
OTarasov-DataArt: Added. There was no unit fixture, the integration tests have been fully commented out since the .NET upgrade, so this adds one. Asserts the run stops on a batch boundary and never advances LastSafelyProcessedDonor, plus a not-cancelled case so the others can't pass vacuously

Atlas.MatchingAlgorithm.Data.Persistent/Migrations/SearchAlgorithmPersistentContextModelSnapshot.cs
OTarasov-DataArt: Not isolated, MatchingAlgorithm.Data and MatchPrediction.Data snapshots are already on 10.0.9, this is the third to catch up. I'd leave it, reverting would just recreate the same diff for the next persistent migration

da-ikovalenko (approved):

Atlas.MatchingAlgorithm/Services/DataRefresh/DataRefreshOrchestrator.cs
OTarasov-DataArt: Yes, essentially. The claim is refused, logged at Info, and the function returns normally, which settles the message, so it isn't redelivered again. The record stays incomplete, and RequestDataRefresh blocks new requests while one exists, so recovery is RunDataRefreshCleanup followed by a fresh request, or manually clearing the lease (documented in README_Support.md). We will raise a separate ticket for a watchdog to automate this safely

da-ikovalenko: Suggestion: What if we build Watchdog on a timer? A scheduled function that periodically looks for refresh records that are unfinished and unleased, and re-publishes a request for each one. The claim guard then decides who actually runs it, and the runner resumes from its last completed stage as it already does.
The pattern already exists in the repo — FinaliseCompletedParallelMatchPredictionRuns does exactly this for match prediction: timer trigger, "awaiting work and not leased" query, per-item lease claim, skip-and-log if someone else got there first.
It heals any kind of stall, not just this one — host restart, deployment mid-run, a cleanup that half-finished. And it doesn't touch the refusal path, so nothing already reviewed changes; the watchdog only publishes, and a duplicate publish is already harmless.
Needs a cron schedule and a grace period longer than the lease duration, so it can't re-request a record whose owner is just between renewals.

PR #1518, opened 2026-09-01
feature: ATL-296: pipeline the donor import stage

OTarasov-DataArt:
This change is [Reviewable](https://reviewable.io/reviews/Anthony-Nolan/Atlas/1518)

### DIFF
--- a/Atlas.MatchingAlgorithm.Test/Services/DataRefresh/DonorImporterTests.cs
+++ b/Atlas.MatchingAlgorithm.Test/Services/DataRefresh/DonorImporterTests.cs
@@ -41,6 +41,14 @@ namespace Atlas.MatchingAlgorithm.Test.Services.DataRefresh
 
         private Fixture fixture;
 
+        /// <summary>Mirrors <c>DonorImporter.BatchSize</c>; tests that care about batch boundaries need to match it.</summary>
+        private const int BatchSize = 10000;
+
+        /// <summary>Generous, because it is only ever reached when the pipeline is broken and the test would hang.</summary>
+        private static readonly TimeSpan PipeliningTimeout = TimeSpan.FromSeconds(10);
+
+        private const string DonorStreamFailureMessage = "the donor stream could not be read";
+
         [SetUp]
         public void SetUp()
         {
@@ -225,11 +233,10 @@ namespace Atlas.MatchingAlgorithm.Test.Services.DataRefresh
         [Test]
         public async Task ImportDonors_WhenCancelledMidImport_StopsOnABatchBoundary()
         {
-            // Two batches' worth of donors, cancelled while the first is being written. The check sits at the top of the
-            // loop, so the first batch completes and the second is never started.
-            const int batchSize = 10000;
+            // Two batches, cancelled while the first is being written. The write side checks the token before beginning
+            // each batch, so the second is never started - even though the read side has likely already buffered it.
             var donor = DonorBuilder.New.With(d => d.AtlasDonorId, fixture.Create<int>()).Build();
-            donorReader.StreamAllDonors().Returns(Enumerable.Repeat(donor, batchSize * 2));
+            donorReader.StreamAllDonors().Returns(Enumerable.Repeat(donor, BatchSize * 2));
 
             var cancellationTokenSource = new CancellationTokenSource();
             donorImportRepository.WhenForAnyArgs(r => r.InsertBatchOfDonors(default)).Do(_ => cancellationTokenSource.Cancel());
@@ -238,6 +245,102 @@ namespace Atlas.MatchingAlgorithm.Test.Services.DataRefresh
                 .ThrowAsync<OperationCanceledException>();
 
             await donorImportRepository.Received(1).InsertBatchOfDonors(Arg.Any<IEnumerable<DonorInfo>>());
+        }
+
+        #endregion
+
+        #region Pipelining
+
+        /// <summary>The write below blocks until the read side reaches the next batch, which run serially it never could.</summary>
+        [Test]
+        public async Task ImportDonors_ReadsTheNextBatchWhileThePreviousOneIsStillBeingWritten()
+        {
+            var secondBatchStarted = new TaskCompletionSource();
+            var donor = DonorBuilder.New.With(d => d.AtlasDonorId, fixture.Create<int>()).Build();
+
+            donorReader.StreamAllDonors().Returns(TwoBatchesSignallingWhenTheSecondIsReached(donor, secondBatchStarted));
+
+            // Captured on the first write only. By the second, the signal is set regardless and proves nothing.
+            bool? readSideRanAheadDuringTheFirstWrite = null;
+            donorImportRepository.WhenForAnyArgs(r => r.InsertBatchOfDonors(default))
+                .Do(_ => readSideRanAheadDuringTheFirstWrite ??= secondBatchStarted.Task.Wait(PipeliningTimeout));
+
+            await donorImporter.ImportDonors();
+
+            readSideRanAheadDuringTheFirstWrite.Should()
+                .BeTrue("the read side must continue fetching donors while a batch is being written, or the stage is still serial");
+        }
+
+        /// <summary>
+        /// The read side now fails on its own thread, and a failure lost there would close the channel cleanly - leaving
+        /// the stage to report success over a partial donor set, silently, for every subsequent search.
+        /// </summary>
+        [Test]
+        public async Task ImportDonors_WhenTheDonorStreamFails_FailsTheImportWithTheUnderlyingFailure()
+        {
+            var donor = DonorBuilder.New.With(d => d.AtlasDonorId, fixture.Create<int>()).Build();
+            donorReader.StreamAllDonors().Returns(DonorStreamThatFailsPartWayThrough(donor));
+
+            var thrown = await donorImporter.Invoking(i => i.ImportDonors()).Should().ThrowAsync<DonorImportHttpException>();
+
+            thrown.WithMessage($"*{DonorStreamFailureMessage}*").WithInnerException<InvalidOperationException>();
+        }
+
+        /// <summary>
+        /// Left running, the read side stays blocked on a full channel holding open the cross-database connection
+        /// <see cref="IDonorReader.StreamAllDonors"/> enumerates, outliving the stage that opened it.
+        /// </summary>
+        [Test]
+        public async Task ImportDonors_WhenAWriteFails_DoesNotReturnUntilTheDonorStreamHasBeenTornDown()
+        {
+            var donorStreamTornDown = new TaskCompletionSource();
+            var donor = DonorBuilder.New.With(d => d.AtlasDonorId, fixture.Create<int>()).Build();
+
+            donorReader.StreamAllDonors().Returns(LongDonorStreamSignallingTeardown(donor, donorStreamTornDown));
+            donorImportRepository.WhenForAnyArgs(r => r.InsertBatchOfDonors(default))
+                .Do(_ => throw new InvalidOperationException("the write failed"));
+
+            await donorImporter.Invoking(i => i.ImportDonors()).Should().ThrowAsync<DonorImportHttpException>();
+
+            donorStreamTornDown.Task.IsCompleted.Should()
+                .BeTrue("the import must not return while a thread is still enumerating the master donor store");
+        }
+
+        private static IEnumerable<Donor> TwoBatchesSignallingWhenTheSecondIsReached(Donor donor, TaskCompletionSource secondBatchStarted)
+        {
+            for (var i = 0; i < BatchSize; i++)
+            {
+                yield return donor;
+            }
+
+            secondBatchStarted.TrySetResult();
+
+            for (var i = 0; i < BatchSize; i++)
+            {
+                yield return donor;
+            }
+        }
+
+        private static IEnumerable<Donor> DonorStreamThatFailsPartWayThrough(Donor donor)
+        {
+            yield return donor;
+            throw new InvalidOperationException(DonorStreamFailureMessage);
+        }
+
+        /// <remarks>Comfortably more donors than the channel can buffer, so the read side is still going when the write fails.</remarks>
+        private static IEnumerable<Donor> LongDonorStreamSignallingTeardown(Donor donor, TaskCompletionSource tornDown)
+        {
+            try
+            {
+                for (var i = 0; i < BatchSize * 20; i++)
+                {
+                    yield return donor;
+                }
+            }
+            finally
+            {
+                tornDown.TrySetResult();
+            }
         }
 
         #endregion
--- a/Atlas.MatchingAlgorithm.Test/Services/DataRefresh/DonorImporterTests.cs
+++ b/Atlas.MatchingAlgorithm.Test/Services/DataRefresh/DonorImporterTests.cs
@@ -1,4 +1,5 @@
 using Atlas.Client.Models.SupportMessages;
+using Atlas.Common.ApplicationInsights;
 using Atlas.DonorImport.ExternalInterface;
 using Atlas.DonorImport.ExternalInterface.Models;
 using Atlas.DonorImport.Test.TestHelpers.Builders.ExternalModels;
@@ -306,6 +307,32 @@ namespace Atlas.MatchingAlgorithm.Test.Services.DataRefresh
                 .BeTrue("the import must not return while a thread is still enumerating the master donor store");
         }
 
+        /// <summary>
+        /// A write failing on its own never observes the channel, so the read side's own failure reaches nobody. Without
+        /// the read side logging it where it is caught, a simultaneous failure of both would leave no trace of the read.
+        /// </summary>
+        [Test]
+        public async Task ImportDonors_WhenBothSidesFailAtOnce_StillRecordsTheReadSideFailure()
+        {
+            var readAboutToFail = new TaskCompletionSource();
+            var donor = DonorBuilder.New.With(d => d.AtlasDonorId, fixture.Create<int>()).Build();
+
+            donorReader.StreamAllDonors().Returns(OneBatchThenFailure(donor, readAboutToFail));
+            donorImportRepository.WhenForAnyArgs(r => r.InsertBatchOfDonors(default)).Do(_ =>
+            {
+                readAboutToFail.Task.Wait(PipeliningTimeout);
+                throw new InvalidOperationException("the write failed");
+            });
+
+            (await donorImporter.Invoking(i => i.ImportDonors()).Should().ThrowAsync<DonorImportHttpException>())
+                .WithMessage("*the write failed*");
+
+            logger.Received().SendTrace(
+                Arg.Is<string>(m => m.Contains(DonorStreamFailureMessage)),
+                LogLevel.Error,
+                Arg.Any<Dictionary<string, string>>());
+        }
+
         private static IEnumerable<Donor> TwoBatchesSignallingWhenTheSecondIsReached(Donor donor, TaskCompletionSource secondBatchStarted)
         {
             for (var i = 0; i < BatchSize; i++)
@@ -324,6 +351,21 @@ namespace Atlas.MatchingAlgorithm.Test.Services.DataRefresh
         private static IEnumerable<Donor> DonorStreamThatFailsPartWayThrough(Donor donor)
         {
             yield return donor;
+            throw new InvalidOperationException(DonorStreamFailureMessage);
+        }
+
+        /// <remarks>
+        /// Yields one whole batch, so the write side gets something to fail on, then signals immediately before failing
+        /// itself - which lets the write side hold off until both failures are genuinely in flight at once.
+        /// </remarks>
+        private static IEnumerable<Donor> OneBatchThenFailure(Donor donor, TaskCompletionSource readAboutToFail)
+        {
+            for (var i = 0; i < BatchSize; i++)
+            {
+                yield return donor;
+            }
+
+            readAboutToFail.TrySetResult();
             throw new InvalidOperationException(DonorStreamFailureMessage);
         }
 
--- a/Atlas.MatchingAlgorithm.Test/Services/DataRefresh/DonorImporterTests.cs
+++ b/Atlas.MatchingAlgorithm.Test/Services/DataRefresh/DonorImporterTests.cs
@@ -329,7 +329,7 @@ namespace Atlas.MatchingAlgorithm.Test.Services.DataRefresh
 
             logger.Received().SendTrace(
                 Arg.Is<string>(m => m.Contains(DonorStreamFailureMessage)),
-                LogLevel.Error,
+                LogLevel.Warn,
                 Arg.Any<Dictionary<string, string>>());
         }
 
--- a/Atlas.MatchingAlgorithm/Services/DataRefresh/DonorImport/DonorImporter.cs
+++ b/Atlas.MatchingAlgorithm/Services/DataRefresh/DonorImport/DonorImporter.cs
@@ -17,6 +17,7 @@ using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Threading;
+using System.Threading.Channels;
 using System.Threading.Tasks;
 using IDonorImportRepository = Atlas.MatchingAlgorithm.Data.Repositories.DonorUpdates.IDonorImportRepository;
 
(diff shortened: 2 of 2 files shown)


Atlas.MatchingAlgorithm/Services/DataRefresh/DonorImport/DonorImporter.cs
```diff
@@ -101,8 +109,149 @@ public async Task ImportDonors(bool shouldMarkDonorsAsUpdated, CancellationToken
...
+                    }
+
+                    // Reified on this thread deliberately: handing over a lazy sequence would move the cost of reading
+                    // those donors onto the write side, which is the serialisation this pipeline exists to remove.
+                    await writer.WriteAsync(donorBatches.Current.ToList(), cancellationToken);
+                }
+
+                writer.TryComplete();
+            }
+            catch (Exception e)
```
da-ikovalenko, line 198: `ReadDonorBatches` always resolves its own exception into `writer.TryComplete(e)` and returns normally, so this method's `Task` can never fault. That makes `AwaitReadTaskQuietly`'s catch block (lines 232-244) unreachable, and silently discards the read-side's root-cause exception whenever the write side has already started propagating a different failure.

Failure scenario: the matching-DB write fails transiently inside `InsertDonorBatch` (line 224, throws directly, bypassing the channel) while the master donor store connection independently drops mid-enumeration on the read thread. Only the write-side exception is ever logged/thrown as `DonorImportHttpException`; the read-side failure is never logged anywhere.

Atlas.MatchingAlgorithm/Services/DataRefresh/DonorImport/DonorImporter.cs
```diff
@@ -38,17 +39,32 @@ public interface IDonorImporter
...
+        /// interrupted import always stops on a batch boundary; batches already buffered by the read side are discarded.
         /// </param>
         Task ImportDonors(bool shouldMarkDonorsAsUpdated = false, CancellationToken cancellationToken = default);
     }
 
     public class DonorImporter : IDonorImporter
     {
         private const int BatchSize = 10000;
+
+        /// <summary>
```
da-ikovalenko, line 52: The `ChannelDepth=3` memory-headroom comment undercounts true peak memory: the read side reifies a batch before it can enter the full channel, and the write side holds its dequeued batch for the duration of `InsertDonorBatch`, so up to 5 reified 10,000-donor batches can be alive at once, not 3.

Risk: a future change that raises `ChannelDepth` based on the comment's implied budget of `ChannelDepth × BatchSize` would actually push peak memory to `(ChannelDepth+2) × BatchSize`, which against the documented ~4.6GB/14GB worker headroom risks an OOM partway through a ~44M-donor refresh.

Atlas.MatchingAlgorithm/Services/DataRefresh/DonorImport/DonorImporter.cs
```diff
@@ -111,9 +260,11 @@ public async Task ImportDonors(bool shouldMarkDonorsAsUpdated, CancellationToken
         /// <returns>Details of donors in the batch that failed import</returns>
         private async Task<IEnumerable<FailedDonorInfo>> InsertDonorBatch(
             List<SearchableDonorInformation> donors,
-            bool shouldMarkDonorsAsUpdated)
+            bool shouldMarkDonorsAsUpdated,
+            int? queueDepth)
         {
-            using (logger.RunTimed($"Import donor batch (BatchSize: {BatchSize})", LogLevel.Verbose))
+            using (logger.RunTimed($"Import donor batch (BatchSize: {BatchSize}, QueueDepth: {queueDepth?.ToString() ?? "unknown"})",
```
da-ikovalenko: This log line interpolates the constant `BatchSize` rather than `donors.Count`, so the final partial batch of any import (donor count not a multiple of 10,000) logs an incorrect batch size. Pre-existing, but this line was touched here (QueueDepth appended) without fixing it, and QueueDepth is meant to help diagnose pipeline starvation using these exact log lines.

da-ikovalenko (commented):
Automated review findings for PR 1518.

One additional finding is not attached inline because it's outside this PR's diff (pre-existing code in `Atlas.Common/ApplicationInsights/Timing/LoggingStopwatch.cs:73`): the shared static `Random` (`RandomSource`) used by `LoggingStopwatch` is not thread-safe, and this PR is the first thing to guarantee `logger.RunTimed` runs concurrently on this call path (read side and write side both construct `LoggingStopwatch` instances concurrently during a data refresh). Concurrent `RandomSource.Next()` calls can race on `Random`'s mutable internal state, corrupting it process-wide (degenerate/duplicate trace IDs for unrelated concurrent operations) or rarely throwing.

Atlas.MatchingAlgorithm/Services/DataRefresh/DonorImport/DonorImporter.cs
```diff
@@ -101,8 +109,149 @@ public async Task ImportDonors(bool shouldMarkDonorsAsUpdated, CancellationToken
...
+            }
+        }
+
+        /// <summary>
+        /// Drives the donor enumerator and hands reified batches to the write side.
+        /// </summary>
+        /// <remarks>
+        /// <see cref="IDonorReader.StreamAllDonors"/> is a synchronous, unbuffered <see cref="IEnumerable{T}"/> over one
+        /// open cross-database connection with no async equivalent, so the enumeration is pushed onto the pool rather
+        /// than the enumerable forced to be async. Expect it to hold a pool thread for most of the stage, not just for
```
da-ikovalenko, line 159: Lower-confidence note: the read side pins one ThreadPool worker thread for most of the stage's ~4+ hour duration, since `StreamAllDonors` is a synchronous, unbuffered `IEnumerable` with no async yield point except when the channel is full. This looks like it's explicitly acknowledged/accepted in this PR's own doc comment, so flagging for visibility rather than as a bug — if this Functions worker needs to grow its thread pool for other concurrent work during the refresh window, one slot is unavailable for the whole run.

da-ikovalenko (changes requested):

Atlas.MatchingAlgorithm/Services/DataRefresh/DonorImport/DonorImporter.cs
OTarasov-DataArt: Correct on both counts. Now logs at the catch site in ReadDonorBatches so the read failure is recorded regardless of who observes the channel (excluding OperationCanceledException, so cancellations don't log as errors). Kept AwaitReadTaskQuietly's catch as defensive and re-commented it. Added a test for this exact scenario.

Atlas.MatchingAlgorithm/Services/DataRefresh/DonorImport/DonorImporter.cs
OTarasov-DataArt: Right, peak is 5 not 3, comment now says (ChannelDepth + 2) * BatchSize. Not an big OOM risk though, in my opinion: measured ~16MB of heap per batch, so the +2 is a constant ~33MB against ~9.4GB of headroom.

Atlas.MatchingAlgorithm/Services/DataRefresh/DonorImport/DonorImporter.cs
OTarasov-DataArt: Fixed, now donors.Count. Fair point

Atlas.MatchingAlgorithm/Services/DataRefresh/DonorImport/DonorImporter.cs
OTarasov-DataArt: Yes, and the mechanism is exactly as you describe. Deliberate and documented in the doc comment. On the conditional, this app only does data refresh, so while one is running there's nothing else to grow the pool for, just health-check pings and a timer tick the incomplete-record guard refuses. If something did arrive, three of the four instantly-available threads are still free

Atlas.MatchingAlgorithm/Services/DataRefresh/DonorImport/DonorImporter.cs
```diff
@@ -101,8 +111,156 @@ public async Task ImportDonors(bool shouldMarkDonorsAsUpdated, CancellationToken
...
+                writer.TryComplete();
+            }
+            catch (Exception e)
+            {
+                // Logged here rather than left to whoever observes the channel. If the write side has already failed on
+                // its own it never reads the completion, so this is otherwise the only record of why the read side
+                // stopped. Cancellation is excluded - losing the lease is expected, and not a failure.
+                if (e is not OperationCanceledException)
+                {
+                    logger.SendTrace($"Donor read failed: {e}", LogLevel.Error);
```
da-ikovalenko: A read-side-only failure is now logged twice at Error level for one incident: this logs `"Donor read failed: {e}"`, then `ImportDonors`'s outer catch (line 109) separately logs `"Donor Import Failed: {ex.Message}"` for the same exception once it propagates back through `AwaitReadTaskQuietly`/`RunImportPipeline`. Worth suppressing one of the two, or tagging them so on-call doesn't see two Error traces for a single failure.

Atlas.MatchingAlgorithm/Services/DataRefresh/DonorImport/DonorImporter.cs
```diff
@@ -101,8 +111,156 @@ public async Task ImportDonors(bool shouldMarkDonorsAsUpdated, CancellationToken
...
+
         /// <param name="donors">Batch of donors to insert into the matching database.</param>
         /// <param name="shouldMarkDonorsAsUpdated"></param>
+        /// <param name="queueDepth">
+        ///     Batches the read side had ready when this write began, or null if the channel cannot report it. Logged
+        ///     beside the write's own duration: a depth sitting at zero means the write side is starved and the
+        ///     pipeline is delivering nothing, which is otherwise indistinguishable from success until the stage as a
+        ///     whole fails to speed up.
+        /// </param>
         /// <param name="batchFetchTime">
```
da-ikovalenko: Pre-existing, but since this doc block was touched in this PR (the `queueDepth` entry added just above): `InsertDonorBatch`'s XML doc still documents a `batchFetchTime` param that doesn't exist on the method's actual signature (`donors`, `shouldMarkDonorsAsUpdated`, `queueDepth`). Worth deleting this stale block while in here.

Atlas.MatchingAlgorithm/Services/DataRefresh/DonorImport/DonorImporter.cs
```diff
@@ -101,8 +111,156 @@ public async Task ImportDonors(bool shouldMarkDonorsAsUpdated, CancellationToken
...
+                    SingleReader = true,
+                    SingleWriter = true,
+                    FullMode = BoundedChannelFullMode.Wait
+                });
+
+            // Linked, so a write-side failure tears the read side down too. Otherwise it would block forever on a full
+            // channel, holding open the cross-database connection it enumerates.
+            using var readCancellation = CancellationTokenSource.CreateLinkedTokenSource(cancellationToken);
+
+            var readTask = Task.Run(() => ReadDonorBatches(batches.Writer, readCancellation.Token), CancellationToken.None);
```
da-ikovalenko, line 137: Low-confidence/minor: `ReadDonorBatches` runs via plain `Task.Run` without `TaskCreationOptions.LongRunning`, so it holds a general-purpose thread-pool thread for most of the multi-hour stage (acknowledged in the method's own doc comment). `Task.Factory.StartNew(..., TaskCreationOptions.LongRunning)` would give it a dedicated thread instead of borrowing from the pool the rest of the worker shares.

da-ikovalenko (commented):
Re-review after the latest commit.

**Jira cross-check (ATL-296 "pipeline the donor import stage 40 — overlap read and write"):** the diff satisfies the ticket's stated approach — bounded `Channel<List<SearchableDonorInformation>>` producer/consumer (`ChannelDepth=3`, within the ticket's 2-4 range), reader kept synchronous on its own thread via `Task.Run` rather than forcing the blocking `IEnumerable` async, failed-donor collection and `SendFailedDonorsAlert` remain on the consumer/main-flow side, no checkpointing added (matches "simpler of the two pipelining tickets"), and cancellation/failure handling across producer and consumer is deliberately worked through with linked `CancellationTokenSource` teardown and dedicated tests. The ticket's "Verify" step (queue-depth counter to distinguish channel-starving from success) is served by the `QueueDepth` value now logged on every batch-insert trace. No gaps against the ticket's requirements.

**Fixed since the last review:** the read-side exception is now logged before being folded into `writer.TryComplete(e)`, the `ChannelDepth` memory comment now correctly documents `(ChannelDepth+2) × BatchSize` peak reified batches, the per-batch log now reports `donors.Count` instead of the constant `BatchSize`, and a new test (`ImportDonors_WhenBothSidesFailAtOnce_StillRecordsTheReadSideFailure`) covers the simultaneous-failure case. Good fixes.

**Still outstanding from the last review (unchanged, out of this PR's diff):** `Atlas.Common/ApplicationInsights/Timing/LoggingStopwatch.cs:73` — the shared static `Random` (`RandomSource`) backing `LoggingStopwatch`'s trace-id generation is not thread-safe, and this PR is the first caller to invoke `logger.RunTimed(...)` from two genuinely concurrent threads (read side and write side) for the full multi-hour duration of the stage. Concurrent `Next()` calls can corrupt `Random`'s internal state, producing degenerate/duplicate ids or, worst case, an out-of-range index throwing from inside the `LoggingStopwatch` constructor mid-import. Worth a follow-up ticket even though it's pre-existing.

New findings below, inline.

Atlas.MatchingAlgorithm/Services/DataRefresh/DonorImport/DonorImporter.cs
OTarasov-DataArt: Fixed, read-side log dropped to Warn. It still clears the configured Info threshold, so a write failure that never observes the channel doesn't bury the read cause, but a read-only failure now raises a single Error from ImportDonors

Atlas.MatchingAlgorithm/Services/DataRefresh/DonorImport/DonorImporter.cs
OTarasov-DataArt: Deleted, thanks

Atlas.MatchingAlgorithm/Services/DataRefresh/DonorImport/DonorImporter.cs
OTarasov-DataArt: I think LongRunning actually might not get us a dedicated thread here. ReadDonorBatches is async, so the non-pool thread only carries it to the first yielding await (WriteAsync when the channel is full); after that the enumeration resumes on the pool and stays there. We'd pay for a thread create/teardown and still hold a pool thread for the rest of the stage. LongRunning really pays off for a genuinely synchronous blocking delegate.
There's also a trap in the switch: Task.Factory.StartNew on a Task-returning lambda gives Task<Task>, which still satisfies AwaitReadTaskQuietly(Task) and compiles clean, so without .Unwrap() the finally stops waiting on the read side at the first suspension. I tried it, the full suite still passes, so we'd have no cover for it.
Leaning toward leaving this as-is, but happy to be corrected.

Atlas.MatchingAlgorithm/Services/DataRefresh/DonorImport/DonorImporter.cs
```diff
@@ -101,19 +111,167 @@ public async Task ImportDonors(bool shouldMarkDonorsAsUpdated, CancellationToken
...
+            ChannelReader<List<SearchableDonorInformation>> reader,
+            bool shouldMarkDonorsAsUpdated,
+            CancellationToken cancellationToken)
+        {
+            var allFailedDonors = new List<FailedDonorInfo>();
+
+            await foreach (var reifiedDonorBatch in reader.ReadAllAsync(cancellationToken))
+            {
+                // Checked here as well as by ReadAllAsync, and this is the check that matters: reading an
+                // already-buffered batch completes without ever consulting the token. Batches behind it are discarded
```
da-ikovalenko: This comment's "discarded rather than drained" claim is only true for an actual cancellation of the token (lease loss) - confirmed by direct verification against the current branch. When the read side instead fails with an exception (e.g. the master-store connection drops) and completes the channel via `writer.TryComplete(e)`, `ReadAllAsync` drains all already-buffered items (via `TryRead`) before the stored exception surfaces on the next iteration - so up to `ChannelDepth` (3) already-read batches (~30k donors) get inserted here *after* the read side has already permanently failed, before the failure ever reaches this loop. Not a correctness bug (the buffered donor data is legitimately read, and a failed stage restarts from scratch per the truncate-log-table invariant elsewhere in this file), but it contradicts this comment's blanket claim and is worth a note - or a matching guard - so a future reader isn't misled about how much work continues after the read side has already died.

Atlas.MatchingAlgorithm/Services/DataRefresh/DonorImport/DonorImporter.cs
```diff
@@ -101,19 +111,167 @@ public async Task ImportDonors(bool shouldMarkDonorsAsUpdated, CancellationToken
...
+            try
+            {
+                return await WriteDonorBatches(batches.Reader, shouldMarkDonorsAsUpdated, cancellationToken);
+            }
+            finally
+            {
+                // Cancel before awaiting: a read side blocked on a full channel has to be released before it can
+                // terminate. Awaiting at all is what stops this method returning, by any path, while a thread is still
+                // enumerating the master donor store.
+                await readCancellation.CancelAsync();
```
da-ikovalenko, line 148: Lower-confidence: `readCancellation.CancelAsync()` here is unguarded, while the very next line (`AwaitReadTaskQuietly`) is deliberately wrapped in try/catch specifically so it can't displace an exception already propagating out of the `try`. `CancellationTokenSource.CancelAsync()` invokes registered callbacks and can itself fault if one throws, which per normal try/finally semantics would replace whatever exception is currently propagating from `WriteDonorBatches`. In practice the only registrations on this linked token are BCL-internal (from `CreateLinkedTokenSource`/pending awaiters in `ReadDonorBatches`), so this is unlikely to trigger - flagging for symmetry with the line below it rather than as a live bug.

da-ikovalenko (commented):
Re-review after the latest commit.

**Both outstanding findings from the last review are fixed:** the read-side failure log is now `LogLevel.Warn` instead of `Error` (with a clear rationale in the comment for why - avoiding a double-Error for one incident while still clearing the Info threshold), and the stale `batchFetchTime` XML doc block on `InsertDonorBatch` has been deleted.

**Jira (ATL-296) status:** unchanged from the last review - the diff still fully satisfies the ticket's stated approach and verification step. No new gaps introduced by this commit.

**Process note:** the automated verification pass for this re-review hit a session rate limit partway through (3 of 6 verification agents failed to complete, and the orchestrating review agent itself was cut off before producing a final consolidated report). I completed the two items below by direct verification against the current PR branch. One additional candidate (that `HlaProcessor`/stage-50 has the same un-pipelined serial batch loop) was also checked directly: it's real, but it's the ticket's own acknowledged separate follow-up (referred to as "stage 50/T2" in ATL-296's description), not a gap in this PR, so it's not raised as a finding here.

Two findings below, inline.

Atlas.MatchingAlgorithm/Services/DataRefresh/DonorImport/DonorImporter.cs
```diff
@@ -101,19 +111,167 @@ public async Task ImportDonors(bool shouldMarkDonorsAsUpdated, CancellationToken
...
+        /// <summary>
+        /// Reads donors from the master store and writes them to the matching database concurrently, rather than
+        /// alternating between the two.
+        /// </summary>
+        /// <remarks>
+        /// The two contend for nothing - the read is network-bound on the master store, the write a bulk insert into a
+        /// different database - so serially the stage costs read + write, and overlapped roughly max(read, write). On a
+        /// 43.9M donor refresh, ~254 minutes against ~128.
+        /// </remarks>
+        private async Task<List<FailedDonorInfo>> RunImportPipeline(bool shouldMarkDonorsAsUpdated, CancellationToken cancellationToken)
```
da-ikovalenko, line 123: Reuse note: this hand-rolled producer/consumer pipeline (bounded `Channel`, linked `CancellationTokenSource` teardown, manual cross-thread exception propagation) is implemented locally to this one class, but the same serial read-batch-then-write-batch shape exists in `Atlas.DonorImport/Services/DonorFileImporter.cs:95` (`foreach (var donorUpdateBatch in donorUpdates.Skip(donorUpdatesToSkip).Batch(BatchSize)) { ...; await donorRecordChangeApplier.ApplyDonorRecordChangeBatch(...); }`). Pipelining that stage later would mean re-implementing this cancellation-linking/teardown/exception-propagation logic (and its subtle races) from scratch instead of reusing a tested, shared primitive. Not asking for an extraction now, just flagging the duplication risk for whoever picks up that follow-up.

Atlas.MatchingAlgorithm.Test/Services/DataRefresh/DonorImporterTests.cs
```diff
@@ -241,5 +249,142 @@ await donorImporter.Invoking(i => i.ImportDonors(false, cancellationTokenSource.
...
+            (await donorImporter.Invoking(i => i.ImportDonors()).Should().ThrowAsync<DonorImportHttpException>())
+                .WithMessage("*the write failed*");
+
+            logger.Received().SendTrace(
+                Arg.Is<string>(m => m.Contains(DonorStreamFailureMessage)),
+                LogLevel.Warn,
+                Arg.Any<Dictionary<string, string>>());
+        }
+
+        private static IEnumerable<Donor> TwoBatchesSignallingWhenTheSecondIsReached(Donor donor, TaskCompletionSource secondBatchStarted)
```
da-ikovalenko, line 336: Minor dedup: `TwoBatchesSignallingWhenTheSecondIsReached`, `DonorStreamThatFailsPartWayThrough`, `OneBatchThenFailure`, and `LongDonorStreamSignallingTeardown` (through line ~385) each hand-code the same "yield N batches, signal, maybe throw" shape, differing mostly in batch count and where the signal/throw happens. A single parameterized helper would replace most of this with less risk of copy-paste drift between them.

da-ikovalenko (commented):
Adding the two minor reuse/dedup findings from the first review that were previously left out (lower priority, included now for the record).

Atlas.MatchingAlgorithm/Services/DataRefresh/DonorImport/DonorImporter.cs
OTarasov-DataArt: Confirmed — reproduced it: fill a depth-3 channel, TryComplete(e), and all three buffered batches come out before the exception surfaces. Comment now covers both paths and notes the drain is harmless, since those donors were read legitimately and a failed stage restarts from scratch.

Atlas.MatchingAlgorithm/Services/DataRefresh/DonorImport/DonorImporter.cs
OTarasov-DataArt: I think we can leave it as-is. The asymmetry is intentional: AwaitReadTaskQuietly is wrapped because the read task routinely carries an exception, whereas the only registrations on this linked token are BCL-internal, so CancelAsync has no failure path to guard. Happy to add it for symmetry if you'd prefer — it's two lines.

Atlas.MatchingAlgorithm.Test/Services/DataRefresh/DonorImporterTests.cs
OTarasov-DataArt: I think I'd leave these. A unified helper is doable for three of the four (the teardown one signals on disposal, which isn't a point in the sequence), but the call sites get worse: DonorStream(donor, BatchSize, tcs, throwAfterSignal: true) instead of OneBatchThenFailure(donor, tcs). The difference between two of these is one yielding a single donor and the other a whole batch, which decides whether the write side runs at all — and that's much easier to miss as an argument than as a name.

da-ikovalenko (commented):
Re-reviewed the latest commit (comment-only change to `WriteDonorBatches`, confirming the discarded-vs-drained asymmetry between cancellation and read-side failure - matches the fix requested in the earlier thread). No functional change, no new findings.

Closed the 6 review threads that were fixed in code: read-side exception logging, `ChannelDepth` memory comment, batch-size log using `donors.Count`, double-logging dropped to `Warn`, stale `batchFetchTime` doc removed, and this latest discarded-vs-drained doc clarification. Left open the 4 threads that were discussed and deliberately left as-is (Task.Run/LongRunning tradeoff, `CancelAsync` symmetry, the pipeline-reuse note with `DonorFileImporter.cs`, and the test-generator dedup) so they stay visible for a final call before merge.

da-ikovalenko (approved):

daria-sorokina-da (approved):
> @daria-sorokina-da reviewed 2 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/Anthony-Nolan/Atlas/1518)* status: all files reviewed, 11 unresolved discussions (waiting on da-ikovalenko and OTarasov-DataArt).

PR #1519, opened 2026-09-03
feature: ATL-303: per-locus inverted index for haplotype survivor filtering

da-ikovalenko:
## Summary
- `CompressedPhenotypeExpander.GetHaplotypesForAllowedLoci` used to scan every haplotype in a HF-set's pool (up to ~300K rows) on every donor comparison. Adds `FrequencySetCacheEntry.AlleleIndex` — a per-locus, per-allele-id inverted index of ascending pool positions, built once per set alongside `ProjectedPool` (eagerly, in `HaplotypeFrequencyCache.BuildEntryFromDatabase`).
- The survivor scan now seeds from the allowed locus with the fewest index candidates (falling back to a full pool scan if the subject is untyped at every allowed locus), instead of testing every pooled haplotype.
- Output is provably unchanged: every true survivor must pass the seed locus's mask too, so the seed candidate list is always a superset of the true survivors, and candidates are visited in original pool order — preserving the insertion order `ExpandedGenotypeTruncater.MostLikelyFirst` uses as its tie-break.

Closes ATL-303 (rescoped 2026-09-02 after PR #1515 fixed the original, larger cost this ticket described).

## Test plan
- [x] `dotnet test Atlas.MatchPrediction.Test` — 284 tests, 0 failures (2 perf tests skipped, as expected/gated)
- [x] New/extended unit tests: `PoolFilterAlleleIdTests` (fully-untyped-at-every-locus fallback, multi-id bucket merge), `FrequencySetCacheEntryTests` (index bucket membership vs. brute-force scan, built-once-per-set reference identity), `ImputationEquivalenceTests` (extended the existing `ProjectedPool` built-once test to also cover `AlleleIndex`)
- [ ] `Atlas.MatchPrediction.Test.Integration`'s `PerformanceBenchmarks.cs` (incl. `RUN_CI_PERF_TESTS=1` for the large-pool cases) — **not run**, this dev environment has neither a local SQL Server instance nor Azurite running. Needs running wherever those are available, to confirm `ShouldHavePercentages` output is byte-identical and to measure the actual speedup.
- [ ] **HLA-matching-expert sign-off** — this touches the clinically-sensitive Match Prediction survivor-filtering step, per `CLAUDE.md`.

- - -
This change is [Reviewable](https://reviewable.io/reviews/Anthony-Nolan/Atlas/1519)


roman-sivak-dataart (approved):

Atlas.MatchPrediction/Services/HaplotypeFrequencies/FrequencySetCacheEntry.cs
```diff
@@ -102,4 +104,64 @@ private DataByResolution<HaplotypeKey[]> ProjectPool()
...
+            valueDqb1: BuildIndexForLocus(haplotypes, Locus.Dqb1, Interner.Dqb1),
+            valueDrb1: BuildIndexForLocus(haplotypes, Locus.Drb1, Interner.Drb1));
+    }
+
+    // Dense, exactly as BuildAllowedAlleleMasks's masks are: AlleleInterner mints ids from 0, so an array indexed by
+    // id needs no hashing to build or to read. Each haplotypes[i] is visited once, in order, so every bucket comes
+    // out ascending for free - which is what lets the caller merge a handful of buckets and stay in pool order.
+    private static int[][] BuildIndexForLocus(HaplotypeKey[] haplotypes, Locus locus, AlleleInterner alleles)
+    {
+        var buckets = new List<int>[alleles.IdCount];
```
SergeySalinAtDataart, line 150: This type is very human-unreadable. Without looking into an IDE I wouldn't have guessed it was an array of Lists of ints? Maybe we can just use List<List<int>> here for readability?

SergeySalinAtDataart (approved):

PR #1517, opened 2026-08-28
feature: ATL-233: genotype conversion performance improvements

SergeySalinAtDataart:
This change is [Reviewable](https://reviewable.io/reviews/Anthony-Nolan/Atlas/1517)

PR #1520, opened 2026-09-03
fix: ATL-337: address PR #1517 review feedback

da-ikovalenko:
## Summary
Follow-up to #1517 (ATL-233: genotype conversion performance improvements), addressing the 10 unresolved review comments tracked in [ATL-337](https://anthonynolan.atlassian.net/browse/ATL-337):

1. **Stale XML doc reference** - `CompressedPhenotypeExpander.cs`'s doc comment now points at `ImputedGenotype.Names`/`.Likelihood` instead of the removed `ImputedGenotypes.GenotypeLikelihoods`.
2. **Swallowed warm failure** - `HaplotypeFrequencyCache.WarmConsolidatedFrequencies` now rethrows after logging when called on the awaited path (`AwaitConsolidatedFrequencyWarm: true`), so `GetOrAddAsync` doesn't cache a half-warmed entry and the next caller retries. Still catches-and-logs on the background path. New tests cover both.
3. **`AwaitConsolidatedFrequencyWarm` shipped disabled** - flipped to `true` in `Atlas.MatchPrediction.Worker/appsettings.json`, matching the setting's own doc comment for this precompute host. Not wired into Terraform (no existing env-var override for this settings section on the Worker container app - flagging that as a separate consideration rather than bundling it here).
4. **Lost type-level derivation guarantee** - `GenotypeAtDesiredResolutions` now takes an `ImputedGenotype` in its constructor (instead of three independent init-only properties), so `HaplotypeResolution`/`GenotypeLikelihood` come from one already-consistent source rather than being set from unrelated literals.
5. **Duplicated null-allele computation** - `GenotypeConverter.ResolveDistinctPGroups` now computes each genotype's null-allele-adjusted form once and returns it (via a new `PGroupResolution` record) for the caller's build loop to reuse, instead of recomputing it a second time per genotype.
6. **Untested summation-order invariant** - `ExpandedGenotypeTruncater.MostLikelyFirst` made `internal` and a new test asserts descending-likelihood order directly on it, since `ImputedGenotypes.Genotypes` is deliberately in expansion order and no longer exposes that invariant.
7. **Hand-rolled interning** - extracted a shared `Interner<TKey>` helper; both `AlleleInterner` and `CompressedPhenotypeExpander.InternHaplotypeNames` now use it instead of two independent dictionary+list implementations.
8. **Test-only helper duplicated in production code** - `LikelihoodsByName` moved from the Test project into `Atlas.MatchPrediction.Models`, so `GenotypeImputationFunctions` and the test suite share one implementation.
9. **Three parallel index-aligned collections** - left as-is. The three-collection shape in `ExpandedGenotypeTruncater` is a deliberate, already-documented LOH-threshold tradeoff (avoiding a large dictionary entry array at the 2,000-genotype cap); collapsing it would need a benchmark to justify, so replying on the review thread instead of changing working code.
10. **Line-length violation** - the 178-char using-alias in `GenotypeConverterTests.cs` now wraps across two lines (C# using-alias directives resolve their target from the global namespace only, so it can't be shortened by relying on the file's other `using` imports).

## Test plan
- [x] `dotnet build --configuration Release` for `Atlas.MatchPrediction`, `.Functions`, `.Worker`, `.Test`, `.Test.Integration`, `.Test.Validation` - all clean, 0 errors
- [x] `dotnet test Atlas.MatchPrediction.Test` - 294 passed, 2 skipped (perf tests), 0 failed
- [x] Verified no other call sites break from the `GenotypeAtDesiredResolutions` constructor change or the moved `LikelihoodsByName` extension

[ATL-337]: https://anthonynolan.atlassian.net/browse/ATL-337?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

- - -
This change is [Reviewable](https://reviewable.io/reviews/Anthony-Nolan/Atlas/1520)

### DIFF
--- a/Atlas.MatchPrediction.Test/Services/HaplotypeFrequencies/HaplotypeFrequencyCacheTests.cs
+++ b/Atlas.MatchPrediction.Test/Services/HaplotypeFrequencies/HaplotypeFrequencyCacheTests.cs
@@ -3,6 +3,7 @@ using System.Collections.Frozen;
 using System.Collections.Generic;
 using System.Linq;
 using System.Threading.Tasks;
+using Atlas.Common.ApplicationInsights;
 using Atlas.Common.Caching;
 using Atlas.Common.Public.Models.GeneticData;
 using Atlas.Common.Test.SharedTestHelpers.Builders;
@@ -15,6 +16,7 @@ using AutoFixture;
 using AwesomeAssertions;
 using Microsoft.Extensions.Options;
 using NSubstitute;
+using NSubstitute.ExceptionExtensions;
 using NUnit.Framework;
 using HfSetHaplotypeNames = Atlas.Common.Public.Models.GeneticData.PhenotypeInfo.LociInfo<string>;
 
@@ -243,6 +245,37 @@ internal class HaplotypeFrequencyCacheTests
         // The standing hazard is a second writer of ConsolidatedFrequencies. Awaiting inside the GetOrAddAsync
         // factory is what prevents it: concurrent callers share one lazy task rather than each starting a warm.
         frequencyConsolidator.Received(1).PreConsolidateFrequenciesForCommonMissingLoci(Arg.Any<FrequencySetCacheEntry>());
+    }
+
+    [Test]
+    public async Task GetAllHaplotypeFrequencies_WhenAwaitingTheWarm_AndPreConsolidationThrows_PropagatesTheFailure()
+    {
+        const int setId = 9;
+        var awaitingSut = BuildSut(awaitConsolidatedFrequencyWarm: true);
+        frequencyRepository.GetAllHaplotypeFrequencies(setId).Returns([Record("a", "b", "c", "dqb1", "drb1", 0.5m)]);
+        frequencyConsolidator.PreConsolidateFrequenciesForCommonMissingLoci(Arg.Any<FrequencySetCacheEntry>())
+            .Throws(new InvalidOperationException("boom"));
+
+        // Awaiting the warm exists precisely so a caller never gets back an entry whose ConsolidatedFrequencies is
+        // silently stuck null - a failed pre-consolidation must surface here, not be swallowed.
+        await awaitingSut.Invoking(c => c.GetAllHaplotypeFrequencies(setId)).Should().ThrowAsync<InvalidOperationException>();
+    }
+
+    [Test]
+    public async Task GetAllHaplotypeFrequencies_WhenNotAwaitingTheWarm_AndPreConsolidationThrows_LogsAndReturnsUnwarmedEntry()
+    {
+        const int setId = 10;
+        frequencyRepository.GetAllHaplotypeFrequencies(setId).Returns([Record("a", "b", "c", "dqb1", "drb1", 0.5m)]);
+        frequencyConsolidator.PreConsolidateFrequenciesForCommonMissingLoci(Arg.Any<FrequencySetCacheEntry>())
+            .Throws(new InvalidOperationException("boom"));
+
+        // The background path has no awaiter to propagate to, so the failure must be caught and logged rather than
+        // faulting an unobserved task - the caller still gets an entry back, just an unwarmed one.
+        var entry = await sut.GetAllHaplotypeFrequencies(setId);
+        await WaitUntil(() => logger.ReceivedCalls().Any());
+
+        entry.ConsolidatedFrequencies.Should().BeNull();
+        logger.Received(1).SendTrace(Arg.Is<string>(msg => msg.Contains("Failed to warm consolidated frequency cache")), LogLevel.Error);
     }
 
     private async Task WaitForConsolidation(int setId) =>
--- a/Atlas.MatchPrediction.Worker/appsettings.json
+++ b/Atlas.MatchPrediction.Worker/appsettings.json
@@ -28,7 +28,7 @@
   },
   "HaplotypeFrequencySetCache": {
     "ActiveSetCacheExpiryMinutes": 5,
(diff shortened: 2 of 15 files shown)


Atlas.MatchPrediction.Test/Services/HaplotypeFrequencies/HaplotypeFrequencyCacheTests.cs
```diff
@@ -245,6 +247,37 @@ public async Task GetAllHaplotypeFrequencies_WhenAwaitingTheWarm_ConsolidatesOnc
...
+    {
+        const int setId = 9;
+        var awaitingSut = BuildSut(awaitConsolidatedFrequencyWarm: true);
+        frequencyRepository.GetAllHaplotypeFrequencies(setId).Returns([Record("a", "b", "c", "dqb1", "drb1", 0.5m)]);
+        frequencyConsolidator.PreConsolidateFrequenciesForCommonMissingLoci(Arg.Any<FrequencySetCacheEntry>())
+            .Throws(new InvalidOperationException("boom"));
+
+        // Awaiting the warm exists precisely so a caller never gets back an entry whose ConsolidatedFrequencies is
+        // silently stuck null - a failed pre-consolidation must surface here, not be swallowed.
+        await awaitingSut.Invoking(c => c.GetAllHaplotypeFrequencies(setId)).Should().ThrowAsync<InvalidOperationException>();
```
SergeySalinAtDataart: Could you please use ```var act = () => sut....
sut.Should().ThrowAsync<>()``` idiom?

Atlas.MatchPrediction/Services/HaplotypeFrequencies/Interner.cs
```diff
@@ -0,0 +1,35 @@
...
+        }
+
+        return id;
+    }
+
+    /// <summary>Query time: never mints new ids. Miss => key not seen by this interner.</summary>
+    public bool TryGetId(TKey key, out int id) => idByKey.TryGetValue(key, out id);
+
+    /// <summary>Dense ids from 0, one entry per distinct key interned, in first-seen order.</summary>
+    public IReadOnlyList<TKey> KeyById => keyById;
```
SergeySalinAtDataart: Can we use a regular list here? Or maybe even make keyById public itself and not reintroduce a property? Otherwise, we would have an interface access which is probably not great. Or maybe make a method like `GetKeyById` instead

Atlas.MatchPrediction.Worker/appsettings.json
```diff
@@ -28,7 +28,7 @@
   },
   "HaplotypeFrequencySetCache": {
     "ActiveSetCacheExpiryMinutes": 5,
-    "AwaitConsolidatedFrequencyWarm": false
+    "AwaitConsolidatedFrequencyWarm": true
```
SergeySalinAtDataart, line 31: Are we going to terraform this?

SergeySalinAtDataart (approved):

Atlas.MatchPrediction.Worker/appsettings.json
da-ikovalenko: Confirmed - and to correct the PR description slightly: this isn't a gap in an existing pattern, there's currently no `HaplotypeFrequencySetCache` Terraform wiring at all on the Worker container app (only the Functions app has `HaplotypeFrequencySetCache:ActiveSetCacheExpiryMinutes` wired, via the `:`-style app settings).

Leaving `AwaitConsolidatedFrequencyWarm` as a checked-in `appsettings.json` default rather than introducing new Terraform wiring here: this flag is a fixed host-behaviour switch (precompute vs. search-serving), not an environment-tunable knob - every environment's Worker should have it `true`, so there's no per-environment value to parameterise. Happy to revisit if we want the flexibility, but wanted to flag it as a separate consideration rather than bundle a new Terraform pattern into this review-feedback PR.

Atlas.MatchPrediction.Test/Services/HaplotypeFrequencies/HaplotypeFrequencyCacheTests.cs
da-ikovalenko: Done - switched to the `var act = () => ...; await act.Should().ThrowAsync<>()` idiom in a886db650.

Atlas.MatchPrediction/Services/HaplotypeFrequencies/Interner.cs
da-ikovalenko: Went with the regular-list option - `KeyById` is now typed as the concrete `List<TKey>` rather than `IReadOnlyList<TKey>`, in a886db650. That resolves the interface-access concern directly at the source (no wrapper property, no new `GetKeyById` method needed): `AlleleInterner.GetName`'s `inner.KeyById[id - 1]` and `IdCount`'s `inner.KeyById.Count` now resolve straight to `List<T>`'s members. Every existing consumer stays source-compatible since `List<T>` is assignable wherever `IReadOnlyList<T>` was expected (e.g. `CompressedPhenotypeExpander.InternHaplotypeNames`'s tuple return type).

OTarasov-DataArt (approved):
> @OTarasov-DataArt reviewed 15 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/Anthony-Nolan/Atlas/1520)* status: all files reviewed, 3 unresolved discussions (waiting on SergeySalinAtDataart).

PR #1521, opened 2026-09-04
feature: ATL-297: pipeline the HLA processing stage

OTarasov-DataArt:
This change is [Reviewable](https://reviewable.io/reviews/Anthony-Nolan/Atlas/1521)

### DIFF
--- a/Atlas.MatchingAlgorithm.Test/Services/DataRefresh/HlaProcessorTests.cs
+++ b/Atlas.MatchingAlgorithm.Test/Services/DataRefresh/HlaProcessorTests.cs
@@ -1,4 +1,4 @@
-using System;
+﻿using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Threading;
@@ -22,9 +22,11 @@ using NUnit.Framework;
 namespace Atlas.MatchingAlgorithm.Test.Services.DataRefresh
 {
     /// <summary>
-    /// Covers the cancellation behaviour of the HLA processing stage. The refresh aborts by cancelling a token when it
-    /// loses its run-level lease, and that abort is only safe if it lands on a batch boundary - stopping mid-batch would
-    /// leave the last-safely-processed donor marker ahead of what was actually written.
+    /// Covers the cancellation and prefetch behaviour of the HLA processing stage. The stage reads pages of donors
+    /// ahead of the processing side, so the two properties that matter are that the resume checkpoint tracks what has
+    /// been *written* rather than what has been read, and that the refresh aborts on a batch boundary when it loses its
+    /// run-level lease - stopping mid-batch, or checkpointing a merely-prefetched page, would leave the
+    /// last-safely-processed donor marker ahead of what was actually written, silently skipping donors on resume.
     /// </summary>
     [TestFixture]
     public class HlaProcessorTests
@@ -36,6 +38,7 @@ namespace Atlas.MatchingAlgorithm.Test.Services.DataRefresh
         private IDonorImportRepository donorImportRepository;
         private IHlaImportRepository hlaImportRepository;
         private IDonorHlaExpander donorHlaExpander;
+        private IMatchingAlgorithmImportLogger logger;
 
         private IHlaProcessor hlaProcessor;
 
@@ -60,8 +63,10 @@ namespace Atlas.MatchingAlgorithm.Test.Services.DataRefresh
             var donorHlaExpanderFactory = Substitute.For<IDonorHlaExpanderFactory>();
             donorHlaExpanderFactory.BuildForSpecifiedHlaNomenclatureVersion(default).ReturnsForAnyArgs(donorHlaExpander);
 
+            logger = Substitute.For<IMatchingAlgorithmImportLogger>();
+
             hlaProcessor = new HlaProcessor(
-                Substitute.For<IMatchingAlgorithmImportLogger>(),
+                logger,
                 donorHlaExpanderFactory,
                 Substitute.For<IHlaMetadataDictionaryFactory>(),
                 Substitute.For<IFailedDonorsNotificationSender>(),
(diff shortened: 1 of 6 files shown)


Atlas.MatchingAlgorithm.Test/Services/DataRefresh/HlaProcessorTests.cs
```diff
@@ -124,13 +129,168 @@ public async Task UpdateDonorHla_WhenNotCancelled_ProcessesEveryBatch()
...
+                {
+                    if (!isFirstWrite)
+                    {
+                        return;
+                    }
+
+                    isFirstWrite = false;
+                    readRanAhead.Task.Wait(TimeSpan.FromSeconds(30))
+                        .Should().BeTrue("the read side should fill the channel while the first batch is being processed");
+                });
```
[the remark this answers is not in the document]
OTarasov-DataArt: Good catch , fixed. Now ReturnsForAnyArgs(async _ => await readRanAhead.Task.WaitAsync(...)), so the processing side awaits the signal instead of holding its thread. I kept the TimeoutException as the failure rather than adding try/catch + Assert.Fail, and moved the explanation into a comment. Also re-checked the test still fails when the pipeline is swapped back for a serial loop, so it isn't passing vacuously now that nothing blocks.

Atlas.MatchingAlgorithm/Services/DataRefresh/HlaProcessing/HlaProcessor.cs
```diff
@@ -169,45 +190,207 @@ private async Task PerformHlaUpdate(
...
+                // subtracting the batchProgress inner timings from the stage's wall clock, so it has to be measured
+                // directly. Each page is already reified by the repository, one query per page, so there is no lazy
+                // work left here to accidentally push back onto the processing side either.
+                await using var donorBatches = batchedDonors.GetAsyncEnumerator(cancellationToken);
+
+                while (true)
                 {
+                    // Checked explicitly: NewOrderedDonorBatchesToImport declares no [EnumeratorCancellation]
+                    // parameter, so the token handed to GetAsyncEnumerator reaches nothing, and nothing else in this
+                    // loop observes one either.
```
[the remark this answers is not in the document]
OTarasov-DataArt: You're right, reworded to say what actually matters: the page read is the one uncancellable step here (no [EnumeratorCancellation] on NewOrderedDonorBatchesToImport), so this check is what stops a fresh page query starting after cancellation was requested

Atlas.MatchingAlgorithm/Services/DataRefresh/HlaProcessing/HlaProcessor.cs
```diff
@@ -169,45 +190,208 @@ private async Task PerformHlaUpdate(
...
+                writer.TryComplete();
+            }
+            catch (Exception e)
+            {
+                // Logged here rather than left to whoever observes the channel. If the processing side has already
+                // failed on its own it never reads the completion, so this is otherwise the only record of why the read
+                // side stopped. Cancellation is excluded - losing the lease is expected, and not a failure.
+                if (e is not OperationCanceledException)
+                {
+                    logger.SendTrace($"Donor read failed during HLA processing: {e}", LogLevel.Error);
```
da-ikovalenko: **Correctness/observability:** a genuine (non-cancellation) read-side failure is now logged twice at two severities. `SendTrace(..., Error)` fires here, then `writer.TryComplete(e)` forwards the same exception unwrapped through `ReadAllAsync` into `ProcessDonorBatches`, which propagates it up to `UpdateDonorHla`'s catch-all, logging it again as `Critical`. Pre-PR this was a single serial loop logging once; now App Insights double-counts the same root cause across Error-trace and Critical-exception telemetry.

Atlas.MatchingAlgorithm/Services/DataRefresh/HlaProcessing/HlaProcessor.cs
```diff
@@ -169,45 +190,208 @@ private async Task PerformHlaUpdate(
...
+                    // the pipeline is delivering nothing - which is otherwise indistinguishable from success until the
+                    // stage as a whole fails to speed up. A healthy read side finds it at capacity.
+                    var queueDepth = batches.Reader.CanCount ? batches.Reader.Count : (int?) null;
+
+                    bool hasNextBatch;
+                    using (logger.RunTimed(
+                        $"{ReadBatchTimingMessage} (QueueDepth: {queueDepth?.ToString() ?? "unknown"})", LogLevel.Verbose))
                     {
-                        continue;
+                        hasNextBatch = await donorBatches.MoveNextAsync();
```
da-ikovalenko: **Behavioural change worth calling out:** the page read (`MoveNextAsync`) and the bulk-insert write (`UpdateDonorBatch`) can now run concurrently against the same transient database for the first time on this stage — previously strictly serial. The "no worse than before" reasoning above (lines 232-237) covers connection lifetime but not concurrent contention on a loaded instance; none of the diff's unit tests exercise real I/O to catch pool exhaustion or slowdown here.

Atlas.MatchingAlgorithm/Services/DataRefresh/HlaProcessing/HlaProcessor.cs
```diff
@@ -169,45 +190,208 @@ private async Task PerformHlaUpdate(
...
+                return await ProcessDonorBatches(
+                    batches.Reader, hlaNomenclatureVersion, updateLastSafelyProcessedDonorId, timerCollection, cancellationToken);
+            }
+            finally
+            {
+                // Cancel before awaiting: a read side blocked on a full channel has to be released before it can
+                // terminate. Awaiting at all is what stops this method returning, by any path, while a page query is
+                // still in flight - otherwise an abandoned read task would go on querying the transient database, and
+                // holding a connection to it, after the stage that owns it has unwound.
+                await readCancellation.CancelAsync();
```
da-ikovalenko: When the process side fails, this cancels the read side, but a page query already mid-`MoveNextAsync` can't observe cancellation (per the comment at 278-281) and always runs to completion before its result is discarded. That's a new wasted DB round-trip introduced by the overlap, and it fires exactly on the already-degraded failure path where extra load is least wanted.

Atlas.MatchingAlgorithm/Services/DataRefresh/HlaProcessing/HlaProcessor.cs
```diff
@@ -169,45 +190,208 @@ private async Task PerformHlaUpdate(
...
+        /// Reads pages of donors out of the transient database while previously-read pages are being expanded and
+        /// written, rather than alternating between the two.
+        /// </summary>
+        /// <remarks>
+        /// The two contend for nothing - the read is a keyset-paged query against the Donors table, the processing is
+        /// HLA expansion plus bulk inserts into the matching HLA tables - so serially the stage costs read +
+        /// processing, and overlapped roughly max(read, processing). Processing is much the larger of the two, so the
+        /// read is expected to disappear behind it almost entirely.
+        /// </remarks>
+        private async Task<List<FailedDonorInfo>> RunHlaProcessingPipeline(
```
da-ikovalenko: **Duplication:** this channel-based prefetch scaffold (linked `CancellationTokenSource`, `TryComplete(e)` protocol, `AwaitReadTaskQuietly` swallow-and-log shape) is nearly identical to the one being introduced for `DonorImporter` on `atl-296-pipeline-donor-import` (referenced by the comment at line 232). Worth considering extracting the shared pattern so a correctness fix to one doesn't silently miss the other.

Atlas.MatchingAlgorithm/Services/DataRefresh/HlaProcessing/HlaProcessor.cs
```diff
@@ -169,45 +190,208 @@ private async Task PerformHlaUpdate(
...
+                    // handed to GetAsyncEnumerator never reaches MoveNextAsync. WriteAsync below does observe it; this
+                    // is what stops a fresh page query being started after cancellation was requested.
                     cancellationToken.ThrowIfCancellationRequested();
 
-                    if (!donorBatch.Any())
+                    // Depth as it stood while this page was being fetched, logged beside the fetch's own duration.
+                    // Sitting at zero means the processing side consumed the previous page the instant it arrived, so
+                    // the pipeline is delivering nothing - which is otherwise indistinguishable from success until the
+                    // stage as a whole fails to speed up. A healthy read side finds it at capacity.
+                    var queueDepth = batches.Reader.CanCount ? batches.Reader.Count : (int?) null;
```
da-ikovalenko: Minor: `queueDepth` and the interpolated log message/stopwatch are built unconditionally on every page read, before `SendTrace` applies the configured `LogLevel` filter. At the default (Info) deployment log level, this allocates on all ~20k page-read traces per full refresh only to discard them. Also, the `CanCount ? Count : null` fallback looks like dead code — `batches` is always a locally-created `Channel.CreateBounded`, whose reader always supports `Count`.

Atlas.MatchingAlgorithm/Services/DataRefresh/HlaProcessing/HlaProcessor.cs
```diff
@@ -169,45 +190,208 @@ private async Task PerformHlaUpdate(
...
+                if (completedDonors.Count >= NumberOfBatchesOverlapOnRestart)
+                {
+                    await updateLastSafelyProcessedDonorId(completedDonors.Peek());
+                }
+            }
+
+            return failedDonors;
+        }
+
+        private async Task AwaitReadTaskQuietly(Task readTask)
```
da-ikovalenko: Minor: this try/catch looks effectively unreachable — `ReadDonorBatches`'s own try/catch already wraps its entire body including `await using` enumerator disposal, so `readTask` shouldn't actually fault under normal operation or the throwing-`Dispose` edge case. Not wrong, just possibly dead defensive code that could mislead a future reader into thinking there's a real, exercised path here.

Atlas.MatchingAlgorithm/Services/DataRefresh/HlaProcessing/HlaProcessor.cs
```diff
@@ -169,45 +190,208 @@ private async Task PerformHlaUpdate(
...
+        {
+            var batches = Channel.CreateBounded<List<DonorInfo>>(
+                new BoundedChannelOptions(ChannelDepth)
+                {
+                    SingleReader = true,
+                    SingleWriter = true,
+                    FullMode = BoundedChannelFullMode.Wait
+                });
+
+            // Linked, so a processing-side failure tears the read side down too. Otherwise it would block forever on a
```
da-ikovalenko: Minor: this comment slightly overstates the linkage's necessity — the unconditional `finally` below (line 251) already cancels `readCancellation` on every exit path (success, cancellation, or failure) regardless of linking. The linkage only buys earlier propagation for the narrower case where the caller's own token cancels before the process side next hits `ThrowIfCancellationRequested` — worth tightening the comment so it doesn't read as a safety requirement.

da-ikovalenko (commented):
Automated review pass (see inline comments). Nothing here blocks merge on its own; the double-logging (line 314) is the one I'd actually want addressed before merge.

One more, not inline because it's outside the diff: `IHlaProcessor.UpdateDonorHla`'s doc comment (line 34, pre-existing, unchanged by this PR) still says cancellation is "observed between batches, never mid-batch," without noting that now depends on the explicit re-check at line 345 (`ProcessDonorBatches`). A future cleanup could read that doc, conclude `ReadAllAsync(cancellationToken)` already covers it, and remove that check as apparently redundant — but per this PR's own comment at 341-344, an already-buffered batch can be yielded without consulting the token. Worth a follow-up to update that doc to point at the real guarantee.

Atlas.MatchingAlgorithm/Services/DataRefresh/HlaProcessing/HlaProcessor.cs
```diff
@@ -60,6 +61,26 @@ public class HlaProcessor : IHlaProcessor
 
         public const int NumberOfBatchesOverlapOnRestart = 3;
 
+        /// <summary>
```
da-ikovalenko, line 66: Minor: this comment reads as slightly self-contradictory — it says depth 1 "would very nearly do," then separately justifies depth 2 via a memory-cap concern that the same comment says doesn't really apply to these raw `DonorInfo` pages. Worth double-checking `ChannelDepth = 2` was deliberately derived rather than carried over without re-verifying against this reasoning.

Atlas.MatchingAlgorithm/Services/DataRefresh/HlaProcessing/HlaProcessor.cs
OTarasov-DataArt: Fixed, matching the approach already reviewed on 296: the cause is traced at Warn and the stage-level Error comes from the propagating exception, so one incident raises one error. Warn still clears the deployed Info threshold, so the cause survives even when the processing side failed first and never observed it. Covered in PrefetchPipelineTests; mutation-checked.

Atlas.MatchingAlgorithm/Services/DataRefresh/HlaProcessing/HlaProcessor.cs
OTarasov-DataArt: Pool exhaustion is ruled out by the code: the read side already held its connection across the yield pre-PR, and the writes always fanned out per locus, so peak connections were ~1 reader + 6 writers before and still are, against a pool of 100. Only the reader's activity changed, not the count.

Query contention on the instance is real but a magnitude question, and no code change is the right answer: the overlap is the deliverable, and depth isn't the lever (at depth 1 a SELECT still runs alongside the inserts). If it does matter, the fix is tuning — depth or pool size, not structural, so nothing to pre-build. Hence documenting it with the read durations as the detector: they'd rise as the pipeline fills rather than staying flat.

Atlas.MatchingAlgorithm/Services/DataRefresh/HlaProcessing/HlaProcessor.cs
OTarasov-DataArt: Fixed: repository method now takes an [EnumeratorCancellation] token, passed to Dapper via CommandDefinition. Comment removed rather than reworded.

Atlas.MatchingAlgorithm/Services/DataRefresh/HlaProcessing/HlaProcessor.cs
OTarasov-DataArt: Extracted to Atlas.Common/Utils/PrefetchPipeline.cs with its own tests. Once 296 is on master I'll rebase and convert DonorImporter onto it here, so both call sites share one copy in this PR

Atlas.MatchingAlgorithm/Services/DataRefresh/HlaProcessing/HlaProcessor.cs
OTarasov-DataArt: Dead CanCount branch is gone, the helper passes an explicit depth probe. Left the interpolation: RunTimed takes an eager string and IAtlasLogger doesn't expose the level, so avoiding it means widening that interface for ~20k short strings across a multi-hour stage.

Atlas.MatchingAlgorithm/Services/DataRefresh/HlaProcessing/HlaProcessor.cs
OTarasov-DataArt: Kept, but the comment now states why: it's awaited from a finally, where a fault would displace the real exception propagating out of the pipeline. Explicitly noted as defensive, not a second reporting path for read errors.

Atlas.MatchingAlgorithm/Services/DataRefresh/HlaProcessing/HlaProcessor.cs
OTarasov-DataArt: Agreed, reworded. The unconditional finally guarantees teardown, linkage only shortens the window.

Atlas.MatchingAlgorithm/Services/DataRefresh/HlaProcessing/HlaProcessor.cs
OTarasov-DataArt: Fair, rewritten. Processing measured ~2× the read, so depth 1 recovers nearly all of it and the second rung only absorbs page-time variance. Dropped the memory framing the comment itself undercut.

da-ikovalenko (approved):

roman-sivak-dataart (approved):

PR #1524, opened 2026-09-08
fix: ATL-305: bound HaplotypeFrequencyCache's per-set cache with LRU eviction

da-ikovalenko:
## Summary
- `HaplotypeFrequencyCache.GetAllHaplotypeFrequencies` cached one `FrequencySetCacheEntry` per haplotype frequency set with no size bound, on a cache shared with the HLA Metadata Dictionary — every distinct set touched within its TTL stayed resident forever. This is a process-wide memory floor that grows independently of concurrency, and one of three compounding causes of Match Prediction ACA Worker OOMs (ATL-303/ATL-304 covered the other two, already fixed by PR #1515). It's now the most urgent of the three, since open PR #1517 wires the Data Refresh precompute path (ATL-117/ATL-233) onto this exact cache.
- Gives the cache its own dedicated instance (`HaplotypeFrequencySetCacheProvider`), isolated from the shared `IPersistentCacheProvider` so a fix here can't force the HLA Metadata Dictionary's caches to also need explicit sizing.
- Bounds it with a new `FrequencySetResidencyTracker` that evicts the **least-recently-used** set once `MaxCachedFrequencySets` is reached, keeping the freshest sets resident. This is driven explicitly rather than left to `MemoryCache`'s own `SizeLimit`, which does **not** evict on overflow — verified empirically (a throwaway repro against the actual `LazyCache`/`MemoryCache` packages) that it rejects the new insert outright instead of displacing an existing entry.
- Adds a dedicated `SetCacheExpiryMinutes` TTL, replacing the previous incidental ~24h default inherited from the shared cache. **Kept, not dropped**, after review: it's currently the only thing bounding staleness of a cached set across a data refresh cycle — neither the frequency-set importer nor `RemoveActiveHaplotypeFrequencySets` invalidate a per-set entry, so without it a stale set could stay resident indefinitely past a refresh that changed or removed it. A live computation already holding a `FrequencySetCacheEntry` reference is unaffected by its cache *key* later expiring - only a subsequent lookup pays a rebuild.
- New settings default conservatively (`MaxCachedFrequencySets: 10`, `SetCacheExpiryMinutes: 60`) everywhere the sibling `ActiveSetCacheExpiryMinutes` setting already lives — ops-tunable via config without a code change.
- **`FrequencySetResidencyTracker` rewritten lock-free** after review: `RecordAccess` runs on every `GetAllHaplotypeFrequencies`/`GetConsolidatedFrequency` call, which is a hot path (once per surviving pooled haplotype during genotype expansion — thousands of times per donor, potentially millions across a search), so the original single global `lock` around every access would have serialized all concurrent donor/patient processing on one mutex. Now uses a `ConcurrentDictionary<int, long>` of last-access ticks (`Interlocked`-incremented, lock-free) for the common "already tracked" path; the eviction scan only runs — under a small dedicated lock — on the rare "just admitted a new distinct set" path, bounded by `capacity`.
- **Tracker kept in sync with the cache's actual contents** after review: added `IFrequencySetResidencyTracker.Forget`, wired via a `RegisterPostEvictionCallback` on the cached entry's `MemoryCacheEntryOptions`, so a TTL expiry (or any other removal) tells the tracker to free that slot instead of continuing to count a phantom entry against `MaxCachedFrequencySets` — which could otherwise cause the tracker to evict a genuinely-resident set to make room for one the cache had already let go.
- Terraform now wired for **all three** hosts (`Atlas.Functions`, `Atlas.MatchPrediction.Functions`, and the Match Prediction ACA Worker), correcting this PR's earlier claim that there was no precedent: the sibling `ActiveSetCacheExpiryMinutes` setting **is** already Terraform-wired for the two Functions apps (`terraform/core/function.tf`, `terraform/core/modules/match_prediction/function.tf`) — only the ACA Worker's `container_app.tf` lacked it (as does the unrelated sibling `AwaitConsolidatedFrequencyWarm`). Since ATL-305 exists specifically to fix ACA Worker OOMs, that's the host most worth making ops-tunable, so new `MATCH_PREDICTION_MAX_CACHED_FREQUENCY_SETS`/`MATCH_PREDICTION_SET_CACHE_EXPIRY_MINUTES` root variables (defaulting to 10/60, matching the app-level defaults) are now threaded through to all three.
- Targets `master` (this branch is only compatible with `master` — it depends on `AwaitConsolidatedFrequencyWarm`/`ProjectedPool` from the ATL-233 feature work already on `master` but not yet on `rc/4.0.0`), for the Atlas 4.1 release.

## Test plan
- [x] `dotnet build` — full solution builds with 0 errors
- [x] `dotnet test Atlas.MatchPrediction.Test` — 302/304 passing (2 skipped perf tests, unrelated), including `FrequencySetResidencyTrackerTests` (LRU eviction, recency protection, repeated eviction, `Forget`) and `HaplotypeFrequencyCacheTests` (cache-level eviction, recency-protection, and TTL/tracker-sync through the real wiring)
- [x] Every other unit test project in the solution (`Atlas.Common.Test`, `Atlas.MatchingAlgorithm.Test`, `Atlas.DonorImport.Test`, etc.) still passes — no regressions
- [ ] `terraform fmt`/`validate` for the new variables/env blocks — no `terraform` CLI available in this environment to run locally; please verify via CI or `/terraform-plan`
- [ ] Manual sanity check with a couple of real HF sets in a local dev DB, per `README_DevelopmentStartUpGuide.md`'s seeding steps (not run as part of this PR — reviewer/QA call)

Risk: low — pure infra/caching change, cached values themselves are unchanged, only retention policy and its implementation change.

- - -
This change is [Reviewable](https://reviewable.io/reviews/Anthony-Nolan/Atlas/1524)

### DIFF
--- a/Atlas.Functions/local.settings.template.json
+++ b/Atlas.Functions/local.settings.template.json
@@ -63,6 +63,8 @@
     "Matching:MessagingServiceBus:SearchResultsTopic": "matching-results-ready",
 
     "HaplotypeFrequencySetCache:ActiveSetCacheExpiryMinutes": 5,
+    "HaplotypeFrequencySetCache:MaxCachedFrequencySets": 10,
+    "HaplotypeFrequencySetCache:SetCacheExpiryMinutes": 60,
 
     "MatchPrediction:GenotypeImputation:MaximumExpandedGenotypesPerInput": 2000,
     "MatchPrediction:AzureStorage:ConnectionString": "UseDevelopmentStorage=true",
--- a/Atlas.MatchPrediction.Functions/local.settings.template.json
+++ b/Atlas.MatchPrediction.Functions/local.settings.template.json
@@ -9,6 +9,8 @@
     "FUNCTIONS_WORKER_RUNTIME": "dotnet-isolated",
     "GenotypeImputation:MaximumExpandedGenotypesPerInput": 2000,
     "HaplotypeFrequencySetCache:ActiveSetCacheExpiryMinutes": 5,
+    "HaplotypeFrequencySetCache:MaxCachedFrequencySets": 10,
+    "HaplotypeFrequencySetCache:SetCacheExpiryMinutes": 60,
     "HlaMetadataDictionary:AzureStorageConnectionString": "UseDevelopmentStorage=true",
     "HlaMetadataDictionary:SearchRelatedMetadata:CacheSlidingExpirationInSeconds": 300,
     "MacDictionary:AzureStorageConnectionString": "UseDevelopmentStorage=true",
--- a/Atlas.MatchPrediction.Worker/appsettings.json
+++ b/Atlas.MatchPrediction.Worker/appsettings.json
@@ -28,7 +28,9 @@
   },
   "HaplotypeFrequencySetCache": {
     "ActiveSetCacheExpiryMinutes": 5,
-    "AwaitConsolidatedFrequencyWarm": false
+    "AwaitConsolidatedFrequencyWarm": false,
+    "MaxCachedFrequencySets": 10,
+    "SetCacheExpiryMinutes": 60
   },
   "GenotypeImputation": {
     "MaximumExpandedGenotypesPerInput": 2000
--- a/Atlas.MatchPrediction/Services/HaplotypeFrequencies/FrequencySetResidencyTracker.cs
+++ b/Atlas.MatchPrediction/Services/HaplotypeFrequencies/FrequencySetResidencyTracker.cs
@@ -1,0 +1,66 @@
+using System;
+using System.Collections.Generic;
+
+namespace Atlas.MatchPrediction.Services.HaplotypeFrequencies;
+
+/// <summary>
+/// Tracks which haplotype frequency set ids are currently resident in <see cref="HaplotypeFrequencyCache"/>'s
+/// dedicated cache, in least-recently-used order, and evicts the least-recently-used set to make room whenever a
+/// not-yet-tracked set would otherwise push residency over <c>capacity</c>.
+///
+/// <para>
+/// This exists because <c>Microsoft.Extensions.Caching.Memory.MemoryCache</c>'s own <c>SizeLimit</c> does not evict
+/// an existing entry to admit a new one once the limit is reached - verified empirically, it rejects the new insert
+/// outright instead, so the newest set would simply never get cached rather than displacing the oldest one. Eviction
+/// of the correct (least-recently-used) entry has to be driven explicitly, from here, one call before the cache is
+/// even asked to store anything.
+/// </para>
+/// </summary>
+internal interface IFrequencySetResidencyTracker
+{
+    /// <summary>
+    /// Records that <paramref name="setId"/> was just accessed: marks it most-recently-used if already tracked, or
+    /// admits it - evicting the least-recently-used tracked set first, if at capacity - if not.
+    /// </summary>
+    void RecordAccess(int setId);
+}
+
+internal sealed class FrequencySetResidencyTracker : IFrequencySetResidencyTracker
+{
+    private readonly int capacity;
+    private readonly Action<int> onEvict;
+    private readonly object lockObject = new();
+
+    // Front = least recently used, back = most recently used.
+    private readonly LinkedList<int> setIdsByRecency = new();
+    private readonly Dictionary<int, LinkedListNode<int>> nodesBySetId = new();
+
+    public FrequencySetResidencyTracker(int capacity, Action<int> onEvict)
+    {
+        this.capacity = capacity;
+        this.onEvict = onEvict;
+    }
+
+    public void RecordAccess(int setId)
+    {
+        lock (lockObject)
+        {
+            if (nodesBySetId.TryGetValue(setId, out var existingNode))
+            {
+                setIdsByRecency.Remove(existingNode);
+                setIdsByRecency.AddLast(existingNode);
+                return;
+            }
+
+            while (nodesBySetId.Count >= capacity)
+            {
+                var leastRecentlyUsed = setIdsByRecency.First!;
+                setIdsByRecency.RemoveFirst();
+                nodesBySetId.Remove(leastRecentlyUsed.Value);
+                onEvict(leastRecentlyUsed.Value);
+            }
+
+            nodesBySetId[setId] = setIdsByRecency.AddLast(setId);
+        }
+    }
+}
--- a/Atlas.MatchPrediction/Services/HaplotypeFrequencies/HaplotypeFrequencyCache.cs
+++ b/Atlas.MatchPrediction/Services/HaplotypeFrequencies/HaplotypeFrequencyCache.cs
@@ -61,9 +61,10 @@ internal class HaplotypeFrequencyCache : IHaplotypeFrequencyCache
 {
     private const string ActiveHaplotypeFrequencySetsCacheKey = "hf-active-sets";
 
-    private static string AllFrequenciesCacheKey(int setId) => $"hf-set-{setId}";
+    internal static string AllFrequenciesCacheKey(int setId) => $"hf-set-{setId}";
 
     private readonly IAppCache cache;
+    private readonly IFrequencySetResidencyTracker residencyTracker;
     private readonly IHaplotypeFrequenciesRepository frequencyRepository;
     private readonly IHaplotypeFrequencySetRepository frequencySetRepository;
     private readonly IFrequencyConsolidator frequencyConsolidator;
@@ -71,14 +72,16 @@ internal class HaplotypeFrequencyCache : IHaplotypeFrequencyCache
     private readonly HaplotypeFrequencySetCacheSettings cacheSettings;
 
     public HaplotypeFrequencyCache(
-        IPersistentCacheProvider persistentCacheProvider,
+        IHaplotypeFrequencySetCacheProvider cacheProvider,
+        IFrequencySetResidencyTracker residencyTracker,
         IHaplotypeFrequenciesRepository frequencyRepository,
         IHaplotypeFrequencySetRepository frequencySetRepository,
         IFrequencyConsolidator frequencyConsolidator,
         IMatchPredictionLogger<MatchProbabilityLoggingContext> logger,
         IOptions<HaplotypeFrequencySetCacheSettings> cacheSettings)
     {
-        cache = persistentCacheProvider.Cache;
+        cache = cacheProvider.Cache;
+        this.residencyTracker = residencyTracker;
         this.frequencyRepository = frequencyRepository;
         this.frequencySetRepository = frequencySetRepository;
         this.frequencyConsolidator = frequencyConsolidator;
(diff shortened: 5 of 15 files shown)


pchenery (approved):
> @pchenery reviewed 15 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/Anthony-Nolan/Atlas/1524)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on da-ikovalenko).

Atlas.Functions/local.settings.template.json
```diff
@@ -63,6 +63,8 @@
     "Matching:MessagingServiceBus:SearchResultsTopic": "matching-results-ready",
 
     "HaplotypeFrequencySetCache:ActiveSetCacheExpiryMinutes": 5,
+    "HaplotypeFrequencySetCache:MaxCachedFrequencySets": 10,
```
SergeySalinAtDataart, line 66: Will we have corresponding terraform settings to back these new up?

Atlas.MatchPrediction.Functions/local.settings.template.json
```diff
@@ -9,6 +9,8 @@
     "FUNCTIONS_WORKER_RUNTIME": "dotnet-isolated",
     "GenotypeImputation:MaximumExpandedGenotypesPerInput": 2000,
     "HaplotypeFrequencySetCache:ActiveSetCacheExpiryMinutes": 5,
+    "HaplotypeFrequencySetCache:MaxCachedFrequencySets": 10,
```
SergeySalinAtDataart, line 12: Will we have corresponding terraform settings to back these new up?

Atlas.MatchPrediction.Worker/appsettings.json
```diff
@@ -28,7 +28,9 @@
   },
   "HaplotypeFrequencySetCache": {
     "ActiveSetCacheExpiryMinutes": 5,
-    "AwaitConsolidatedFrequencyWarm": false
+    "AwaitConsolidatedFrequencyWarm": false,
```
SergeySalinAtDataart, line 31: Will we have corresponding terraform settings to back these new up?

Atlas.MatchPrediction/Services/HaplotypeFrequencies/HaplotypeFrequencyCache.cs
```diff
@@ -115,6 +118,11 @@ public HaplotypeFrequencyCache(
     /// <inheritdoc />
     public async Task<FrequencySetCacheEntry> GetAllHaplotypeFrequencies(int setId)
     {
+        // Before the cache is even asked to store anything: marks setId most-recently-used, evicting the least-
+        // recently-used tracked set first if admitting a not-yet-tracked setId would exceed MaxCachedFrequencySets.
+        // See FrequencySetResidencyTracker's doc comment for why eviction has to be driven explicitly like this.
+        residencyTracker.RecordAccess(setId);
```
SergeySalinAtDataart, line 124: This is quite a hot path of our application and `residencyTracker.RecordAccess` introduces a lock there. I think it might incur a performance hit

Atlas.MatchPrediction/Services/HaplotypeFrequencies/FrequencySetResidencyTracker.cs
```diff
@@ -0,0 +1,66 @@
...
+{
+    private readonly int capacity;
+    private readonly Action<int> onEvict;
+    private readonly object lockObject = new();
+
+    // Front = least recently used, back = most recently used.
+    private readonly LinkedList<int> setIdsByRecency = new();
+    private readonly Dictionary<int, LinkedListNode<int>> nodesBySetId = new();
+
+    public FrequencySetResidencyTracker(int capacity, Action<int> onEvict)
```
SergeySalinAtDataart, line 55: I think we might have a problem with sets inside tracker drifting from sets inside the cache. Maybe we can somehow make them communicate so their knowledge about HF sets is in sync with each other so that when TTL removes a set from cache it is also evicted from here?

Atlas.MatchPrediction/Services/HaplotypeFrequencies/HaplotypeFrequencyCache.cs
```diff
@@ -139,6 +147,10 @@ public async Task<FrequencySetCacheEntry> GetAllHaplotypeFrequencies(int setId)
                 }
 
                 return entry;
+            },
+            new MemoryCacheEntryOptions
+            {
+                AbsoluteExpirationRelativeToNow = TimeSpan.FromMinutes(cacheSettings.SetCacheExpiryMinutes)
```
SergeySalinAtDataart, line 153: Are we benefitting from the absolute expiration? What if we drop it during the hot computation relying on the exact HF set? I understand that probability is low, but still, we'll have to load everything from scratch and during that we'll have to wait for cache to load and then to warm. I'm not sure this is the right option. Opinions?

README_MatchPredictionAlgorithm.md
```diff
@@ -202,3 +202,22 @@ The pre-consolidation runs either way, and there is exactly one writer of `Frequ
 in both modes — the setting moves that work on to or off the critical path, nothing more. Note that a machine with spare
 cores will show `true` as a wall-clock regression, because the warm no longer overlaps the first subjects.
 
+### `HaplotypeFrequencySetCache.MaxCachedFrequencySets` / `HaplotypeFrequencySetCache.SetCacheExpiryMinutes`
+
+Bound the size of `HaplotypeFrequencyCache`'s dedicated cache of per-set frequency data (`FrequencySetCacheEntry`),
+which otherwise grows unbounded: every distinct haplotype frequency set (registry/ethnicity) touched within a rolling
```
SergeySalinAtDataart: Technically, incorrect: memory is bound by the total active sets count (currently, 208, totalling about 3,200 MiB)

SergeySalinAtDataart (changes requested):

Atlas.Functions/local.settings.template.json
da-ikovalenko: Good catch — the PR description's claim was wrong for this host: `ActiveSetCacheExpiryMinutes` (this setting's sibling) is already Terraform-wired here (`terraform/core/function.tf:98`), so there was precedent I'd missed. Wired `MaxCachedFrequencySets`/`SetCacheExpiryMinutes` into Terraform for this app (and `Atlas.MatchPrediction.Functions`/the ACA Worker) in dd79bbc61 — new root variables `MATCH_PREDICTION_MAX_CACHED_FREQUENCY_SETS`/`MATCH_PREDICTION_SET_CACHE_EXPIRY_MINUTES`, defaulting to 10/60 to match the app-level defaults.

Atlas.MatchPrediction.Functions/local.settings.template.json
da-ikovalenko: Same as the `Atlas.Functions` comment — `ActiveSetCacheExpiryMinutes` is already Terraform-wired here too (`terraform/core/modules/match_prediction/function.tf:26`), so there was precedent for this host. Wired `MaxCachedFrequencySets`/`SetCacheExpiryMinutes` in dd79bbc61 via new module variables `MAX_CACHED_FREQUENCY_SETS`/`SET_CACHE_EXPIRY_MINUTES`.

Atlas.MatchPrediction.Worker/appsettings.json
da-ikovalenko: This one genuinely had no precedent — `ActiveSetCacheExpiryMinutes` isn't wired in the ACA Worker's `container_app.tf` either, same as the sibling `AwaitConsolidatedFrequencyWarm`. But this is the host ATL-305 exists to fix, so leaving it config-only would miss the point of ops-tunability. Wired both new settings here too in dd79bbc61, as two new `env` blocks (`HaplotypeFrequencySetCache__MaxCachedFrequencySets`/`HaplotypeFrequencySetCache__SetCacheExpiryMinutes`) fed from the same new root variables.

Atlas.MatchPrediction/Services/HaplotypeFrequencies/HaplotypeFrequencyCache.cs
da-ikovalenko: Confirmed, and the concern was real, not just theoretical: `RecordAccess` also runs on every `GetConsolidatedFrequency` call (it re-enters `GetAllHaplotypeFrequencies`), which is invoked once per surviving pooled haplotype during genotype expansion — thousands of times per donor, potentially millions across a search. Rewrote `FrequencySetResidencyTracker` in dd79bbc61 to use a `ConcurrentDictionary<int, long>` of last-access ticks (lock-free `AddOrUpdate` for the common 'already tracked' case) instead of a `lock` + `LinkedList`. The only remaining lock guards the eviction scan, which only runs on the rare 'just admitted a new distinct set' path, not on every hot-path read.

Atlas.MatchPrediction/Services/HaplotypeFrequencies/FrequencySetResidencyTracker.cs
da-ikovalenko: Agreed, this was a real gap. Added `IFrequencySetResidencyTracker.Forget` and wired it through a `RegisterPostEvictionCallback` on the `MemoryCacheEntryOptions` used for the per-set cache entry (dd79bbc61), so any removal — TTL expiry or an explicit `Remove` — tells the tracker to free that slot instead of continuing to count a phantom entry against `MaxCachedFrequencySets`. Added a new test (`GetAllHaplotypeFrequencies_AfterCacheEntryExpiresIndependently_DoesNotEvictAStillResidentSetInItsPlace`) that reproduces exactly the failure mode you're describing: without the fix, a TTL-expired-but-still-tracked entry could cause the tracker to evict a genuinely resident set in its place. One implementation note surfaced while writing that test: `MemoryCache` dispatches post-eviction callbacks on the thread pool rather than synchronously with `Remove`, so there's a small window of eventual (not immediate) consistency — fine here, since it only matters relative to a 60-minute TTL.

Atlas.MatchPrediction/Services/HaplotypeFrequencies/HaplotypeFrequencyCache.cs
da-ikovalenko: Looked into this — keeping `SetCacheExpiryMinutes` is the right call. It's currently the *only* thing bounding staleness of a cached set across a data refresh cycle: neither the frequency-set importer nor `RemoveActiveHaplotypeFrequencySets` invalidate a per-set `hf-set-{setId}` entry, so without a TTL a resident set could keep serving frequency data indefinitely past a refresh that changed or removed it. On the specific worry — a live computation losing its warm set mid-computation — that doesn't actually happen: a caller that's already been handed a `FrequencySetCacheEntry` reference holds that object directly, so removing the cache *key* later doesn't touch it; only a *new* lookup after expiry pays a rebuild. Combined with the tracker-sync fix above (dd79bbc61), this is now a fully consistent tradeoff. Agreed it'd be nicer to invalidate explicitly from the data-refresh path instead of via a timer — think that's worth a small follow-up ticket, but out of scope here.

README_MatchPredictionAlgorithm.md
da-ikovalenko: Fair, that was imprecise — fixed in dd79bbc61 to say it's bounded by the number of active sets (currently 208, ~3,200 MiB) rather than literally unbounded, while keeping the point that it's still a process-wide memory floor that grows independently of concurrency/request volume.

SergeySalinAtDataart (approved):

PR #1525, opened 2026-09-08
feature: ATL-298: reuse SqlBulkCopy per destination table during data refresh

OTarasov-DataArt:
This change is [Reviewable](https://reviewable.io/reviews/Anthony-Nolan/Atlas/1525)

### DIFF
--- a/Atlas.MatchingAlgorithm.Data/Repositories/DonorUpdates/DonorUpdateRepositoryBase.cs
+++ b/Atlas.MatchingAlgorithm.Data/Repositories/DonorUpdates/DonorUpdateRepositoryBase.cs
@@ -1,7 +1,9 @@
+using System;
 using System.Collections.Generic;
 using System.Data;
 using System.Linq;
 using System.Threading.Tasks;
+using System.Transactions;
 using Atlas.Common.ApplicationInsights;
 using Atlas.Common.ApplicationInsights.Timing;
 using Atlas.Common.Public.Models.GeneticData;
@@ -23,6 +25,22 @@ namespace Atlas.MatchingAlgorithm.Data.Repositories.DonorUpdates
     public abstract class DonorUpdateRepositoryBase : Repository
     {
         protected readonly IAtlasLogger logger;
+
+        private const string DonorsTableName = "Donors";
+
+        private const int DefaultBulkInsertTimeoutInSeconds = 3600;
+
+        /// <summary>
+        /// The matching HLA tables are the largest writes in the system, hence a far longer timeout than
+        /// <see cref="DefaultBulkInsertTimeoutInSeconds"/>. Shared with <see cref="BuildReusableBulkCopies"/>, which
+        /// pre-builds those tables' bulk copies, so that the two cannot drift apart.
+        /// </summary>
+        private const int MatchingHlaBulkInsertTimeoutInSeconds = 14400;
+
+        /// <summary>
+        /// Non-null only between <see cref="OpenBulkWriteSession"/> and the disposal of what it returned.
+        /// </summary>
+        private BulkCopySession activeBulkCopySession;
 
         // The order of these matters when setting up the datatable - if re-ordering, also re-order datatable contents
         private readonly string[] donorInsertDataTableColumnNames =
@@ -61,6 +79,30 @@ namespace Atlas.MatchingAlgorithm.Data.Repositories.DonorUpdates
             this.logger = logger;
         }
 
+        /// <summary>
+        /// Opens a session over which this repository's bulk copies are reused, rather than rebuilt per write.
+        /// Dispose the returned handle to close it.
+        /// </summary>
+        /// <remarks>
+        /// A <see cref="SqlBulkCopy"/> asks the server to describe its destination table before sending any rows, and
+        /// that description is cached per instance, so rebuilding one per write pays for a round trip that only ever
+        /// returns the same answer. Over a full data refresh that came to over an hour of waiting on the destination
+        /// metadata catalogue queries for a few minutes of server time.
+        ///
+        /// Not every write can take part - see <see cref="GetReusableBulkCopy"/> - so this is an optimisation that a
+        /// caller opts into for a whole stage, not a change to how any individual write behaves.
+        /// </remarks>
+        public IDisposable OpenBulkWriteSession()
+        {
+            if (activeBulkCopySession != null)
+            {
+                throw new InvalidOperationException($"A bulk write session is already open on this {GetType().Name}.");
+            }
+
+            activeBulkCopySession = new BulkCopySession(this);
+            return activeBulkCopySession;
+        }
+
         public async Task InsertBatchOfDonors(IEnumerable<DonorInfo> donors)
         {
             var donorInfos = donors.ToList();
@@ -72,7 +114,7 @@ namespace Atlas.MatchingAlgorithm.Data.Repositories.DonorUpdates
 
             var dataTable = BuildDonorInsertDataTable(donorInfos);
 
-            await BulkInsertDataTable("Donors", dataTable, donorInsertDataTableColumnNames);
+            await BulkInsertDataTable(DonorsTableName, dataTable, donorInsertDataTableColumnNames);
         }
 
         public async Task AddMatchingRelationsForExistingDonorBatch(
@@ -183,7 +225,15 @@ namespace Atlas.MatchingAlgorithm.Data.Repositories.DonorUpdates
             var dataTable = BuildPerLocusPGroupDataTable(donors, locus, timerCollection);
             buildDataTableTimer?.Dispose();
 
-            using (var transactionScope = new AsyncTransactionScope())
+            // The scope is here to make the DELETE below and the insert that follows it one atomic unit. On the create
+            // path there is no DELETE to be atomic with, and the insert is already atomic in its own right: the
+            // UseInternalTransaction option commits per bulk-copy batch, and one donor batch's rows at one locus
+            // (2 * HlaProcessor.BatchSize at the very most) never reach the 10,000 row batch size.
+            //
+            // Skipping it there is also what leaves the reusable bulk copies usable - see GetReusableBulkCopy. Where a
+            // caller asked for the whole upsert to be transactional, the outer scope opened by
+            // UpsertMatchingPGroupsAtSpecifiedLoci is ambient regardless, and this scope only ever joined it.
+            using (var transactionScope = new OptionalAsyncTransactionScope(!isKnownToBeCreate))
             {
                 using (timerCollection?.TimeInnerOperation(DataRefreshTimingKeys.HlaUpsert_BulkInsertSetup_DeleteExistingRecords_TimerKey))
                 {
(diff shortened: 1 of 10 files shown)


Atlas.MatchingAlgorithm.Data/Repositories/DonorUpdates/DonorUpdateRepositoryBase.cs
```diff
@@ -183,7 +225,15 @@ private async Task UpsertMatchingPGroupsAtLocus(
...
-            using (var transactionScope = new AsyncTransactionScope())
+            // The scope is here to make the DELETE below and the insert that follows it one atomic unit. On the create
+            // path there is no DELETE to be atomic with, and the insert is already atomic in its own right: the
+            // UseInternalTransaction option commits per bulk-copy batch, and one donor batch's rows at one locus
+            // (2 * HlaProcessor.BatchSize at the very most) never reach the 10,000 row batch size.
+            //
+            // Skipping it there is also what leaves the reusable bulk copies usable - see GetReusableBulkCopy. Where a
+            // caller asked for the whole upsert to be transactional, the outer scope opened by
+            // UpsertMatchingPGroupsAtSpecifiedLoci is ambient regardless, and this scope only ever joined it.
+            using (var transactionScope = new OptionalAsyncTransactionScope(!isKnownToBeCreate))
```
da-ikovalenko: Worth calling out explicitly (in the PR description or a comment) that this asymmetry has two knock-on effects beyond what's documented here:

1. **Donor updates get none of the SqlBulkCopy-reuse speedup.** The update path (`isKnownToBeCreate == false`) always keeps a real ambient transaction here (needed for DELETE+INSERT atomicity), so `GetReusableBulkCopy` always sees `Transaction.Current != null` and always falls back to building a fresh `SqlBulkCopy`. Only donor *creation* benefits from ATL-298's optimisation.
2. **`DataRefreshDonorUpdatesShouldBeFullyTransactional` silently disables the whole feature.** When that setting is on, `runAllHlaInsertionsInASingleTransactionScope` wraps the *entire* batch (all loci, creates and updates alike) in one outer real `TransactionScope` in `UpsertMatchingPGroupsAtSpecifiedLoci`, regardless of `isKnownToBeCreate`. `Transaction.Current` is then never null anywhere in the batch, so every write in that mode falls back to a fresh `SqlBulkCopy` — the whole reuse optimisation becomes a no-op for the whole refresh, with nothing in logs/telemetry to reveal it. An environment running with that flag on still pays the session open/close cost for zero benefit.

Neither is a bug given the correctness constraints, but both seem worth a line in the class remarks (or a log/metric) so a future perf investigation doesn't have to rediscover this.

Atlas.MatchingAlgorithm.Data/Repositories/DonorUpdates/DonorUpdateRepositoryBase.cs
```diff
@@ -303,26 +353,66 @@ protected DataTable BuildPerLocusPGroupDataTable(
...
+        /// an ambient transaction keep building an instance of their own, and behave as they did before sessions existed.
+        ///
+        /// The silent path is the dangerous one, and it was measured rather than assumed: reusing an instance
+        /// regardless of the ambient transaction, and then rolling that transaction back, leaves the rows written - so
+        /// a run with DataRefreshDonorUpdatesShouldBeFullyTransactional set would lose transactionality from its
+        /// second batch onwards, with nothing in the logs to say so. That is what
+        /// InsertBatchOfDonors_WithinOneBulkWriteSession_UnderATransactionThatRollsBack_WritesNothing pins down, and
+        /// it is the only test that fails if the check below is dropped.
+        /// </remarks>
+        private SqlBulkCopy GetReusableBulkCopy(string tableName) =>
```
da-ikovalenko, line 425: The safety argument a few lines up ("one donor batch's rows at one locus never reach the 10,000 row batch size") depends on `HlaProcessor.BatchSize` staying under half of the internal-transaction batch size, but that's only comment-enforced here — nothing ties the two constants together in code.

If `HlaProcessor.BatchSize` is later raised (e.g. for throughput) without noticing this coupling, a create-path bulk insert could span multiple internally-committed sub-batches with no wrapping transaction to make the whole insert atomic on partial failure, and nothing would fail loudly to catch it.

Worth either asserting the relationship at startup (e.g. `Debug.Assert` or a settings validator: `HlaProcessor.BatchSize * 2 < 10_000`), or referencing `BatchSize` directly instead of the literal `10000` so a change to one is visible at the other.

Atlas.MatchingAlgorithm.Data/Repositories/DonorUpdates/DonorUpdateRepositoryBase.cs
```diff
@@ -338,6 +428,67 @@ private SqlBulkCopy BuildSqlBulkCopy(string tableName, string[] columnNames, int
...
+        /// </summary>
+        /// <remarks>
+        /// Eagerly, because a <see cref="SqlBulkCopy"/> costs nothing until its first write - it does not even open its
+        /// connection - and because a cache populated up front needs no synchronising against the per-locus writes,
+        /// which run concurrently. One instance per table also means no instance is ever shared between those writes:
+        /// they go to one table each, and a <see cref="SqlBulkCopy"/> rejects a second concurrent write on the same
+        /// instance. Nor do successive batches overlap, since <see cref="UpsertMatchingPGroupsAtSpecifiedLoci"/> awaits
+        /// every locus' write before it returns.
+        /// </remarks>
+        private IReadOnlyDictionary<string, SqlBulkCopy> BuildReusableBulkCopies()
```
da-ikovalenko, line 468: Minor: this always builds all 6 bulk copies (Donors + 5 matching-HLA loci) regardless of which subset the calling stage actually writes to — `DonorImporter`'s session only ever writes `Donors`, and `HlaProcessor`'s only ever writes the 5 locus tables. Not a correctness issue (a `SqlBulkCopy` costs nothing until its first write, per the remarks above), just some avoidable dead setup/teardown work every refresh run. Not blocking, just flagging in case it's easy to scope the built set to what the caller actually needs.

Atlas.MatchingAlgorithm.Data/Repositories/DonorUpdates/DonorUpdateRepositoryBase.cs
```diff
@@ -338,6 +428,67 @@ private SqlBulkCopy BuildSqlBulkCopy(string tableName, string[] columnNames, int
...
+            {
+                this.repository = repository;
+                bulkCopiesByTableName = repository.BuildReusableBulkCopies();
+            }
+
+            /// <returns>This session's bulk copy for the table, or null if the session does not cover that table.</returns>
+            internal SqlBulkCopy BulkCopyFor(string tableName) =>
+                bulkCopiesByTableName.TryGetValue(tableName, out var bulkCopy) ? bulkCopy : null;
+
+            public void Dispose()
```
da-ikovalenko: `Dispose()` here isn't idempotent — a second call would re-iterate and re-dispose the already-closed `SqlBulkCopy` instances and re-null `activeBulkCopySession`. `Atlas.Common.Utils.Disposable.DisposableAction` already exists elsewhere in the codebase specifically to guard against double-dispose (tracks a `hasDisposed` flag). Might be worth reusing that here, or adding the same guard, so this doesn't silently diverge if double-dispose safety is hardened elsewhere later. Low priority since nothing in the current call sites disposes twice.

da-ikovalenko (commented):
Reviewed the `TransactionScope` change and the bulk-copy reuse mechanism it enables. Left a few inline notes on behavioural side-effects worth confirming are intentional.

Atlas.MatchingAlgorithm.Data/Repositories/DonorUpdates/DonorUpdateRepositoryBase.cs
OTarasov-DataArt: Both added to the class remarks, and you were right that a comment does nothing for a log reader, so I've added the trace too. Closing a session now emits one trace with a ReusedWrites count, so a session that reused nothing says so outright. The count only increments where reuse actually happened, so the fully-transactional case reports 0 rather than looking like it worked. Two unit tests and one integration test cover it. Small addition on your first point: the update path doesn't just fall back, it never opens a session at all, OpenBulkWriteSession is only on IDonorImportRepository, and DonorUpdateRepository implements IDonorUpdateRepository. So donor creation really is the only beneficiary.

Atlas.MatchingAlgorithm.Data/Repositories/DonorUpdates/DonorUpdateRepositoryBase.cs
OTarasov-DataArt: Fixed, though not as suggested, referencing HlaProcessor.BatchSize here would invert the layering, and Debug.Assert wouldn't fire in Release. Instead the check reads the rows in hand: isKnownToBeCreate && dataTable.Rows.Count <= BulkCopyBatchSize (10000, now a named constant), so there's nothing left to keep in sync. Verified the fallback rather than assuming it: 5,001 records, over the threshold at all five loci, and the write still succeeds atomically in 349ms with no distributed promotion. Also worth knowing: DonorImporter.BatchSize is already 10,000, exactly at the boundary; harmless today since that path never had a scope, but it's the same coupling with no margin.

Atlas.MatchingAlgorithm.Data/Repositories/DonorUpdates/DonorUpdateRepositoryBase.cs
OTarasov-DataArt: Agreed it's dead work. Scoping it means either leaking table names into the service layer or splitting this in two, for ~5 allocations and disposals per stage on objects that never reach the database, so I've recorded the reasoning in the remarks instead. Happy to split it if you'd prefer.

Atlas.MatchingAlgorithm.Data/Repositories/DonorUpdates/DonorUpdateRepositoryBase.cs
OTarasov-DataArt: Fixed, BulkCopySession now exposes Close() and OpenBulkWriteSession returns new DisposableAction(session.Close), reusing the existing guard rather than duplicating one, with a test that a double dispose reports once. The part that mattered was activeBulkCopySession = null: re-disposing the SqlBulkCopy instances is harmless, but a second Close would clear whichever session is registered by then, orphaning a newer one's bulk copies and their connections. Close is still only correct because of the wrapper, can add if (repository.activeBulkCopySession == this) to make it correct standalone if you want.

Atlas.MatchingAlgorithm.Data/Repositories/DonorUpdates/DonorUpdateRepositoryBase.cs
da-ikovalenko: Agreed, the documentation is enough here - no need to split the method for this. Resolving.

da-ikovalenko (approved):

daria-sorokina-da (approved):
> @daria-sorokina-da reviewed 10 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/Anthony-Nolan/Atlas/1525)* status: all files reviewed, 4 unresolved discussions (waiting on da-ikovalenko and OTarasov-DataArt).

PR #1528, opened 2026-09-16
feature: ATL-320: add watchdog to recover stalled data refreshes

OTarasov-DataArt:
This change is [Reviewable](https://reviewable.io/reviews/Anthony-Nolan/Atlas/1528)


Atlas.MatchingAlgorithm.Functions.DataRefresh/Functions/DataRefreshFunctions.cs
```diff
@@ -64,6 +67,18 @@ public async Task SubmitDataRefreshRequest([TimerTrigger("%DataRefresh:CronTab%"
             await dataRefreshRequester.RequestDataRefresh(request, false);
         }
 
+        /// <summary>
+        /// Re-requests every data refresh that has stalled, i.e. that is still open but that nothing is working on.
+/// Such a record blocks all subsequent refreshes until it is completed; the watchdog re-requests it automatically.
+        /// </summary>
+        [SuppressMessage(null, SuppressMessage.UnusedParameter, Justification = SuppressMessage.UsedByAzureTrigger)]
```
SergeySalinAtDataart: We don't need to use [SuppressMessage(...)], it's obsolete legacy of the Elders. If you wish, you can replace the `timerInfo` with `_` (underscore) to tell both users and the compiler that the parameter is unused

Atlas.MatchingAlgorithm.Test/Services/DataRefresh/DataRefreshWatchdogTests.cs
```diff
@@ -0,0 +1,242 @@
...
+            GivenStalledRecords(StalledRecordId);
+
+            await dataRefreshWatchdog.RecoverStalledRefreshes();
+
+            await serviceBusClient.Received(1).PublishToRequestTopic(
+                Arg.Is<ValidatedDataRefreshRequest>(r => r.DataRefreshRecordId == StalledRecordId));
+        }
+
+        [Test]
+        public async Task RecoverStalledRefreshes_WithMultipleStalledRecords_ReRequestsAllOfThem()
```
SergeySalinAtDataart: I think this test supercedes the previous one and the previous one `RecoverStalledRefreshes_ReRequestsEachStalledRecord` should be removed

Atlas.MatchingAlgorithm.Test/Services/DataRefresh/DataRefreshWatchdogTests.cs
```diff
@@ -0,0 +1,242 @@
...
+        [Test]
+        public async Task RecoverStalledRefreshes_WhenOneRecordFails_StillReRequestsTheOthers()
+        {
+            GivenStalledRecords(1, 2, 3);
+            serviceBusClient.PublishToRequestTopic(Arg.Is<ValidatedDataRefreshRequest>(r => r.DataRefreshRecordId == 2))
+                .ThrowsAsync(new Exception("bus is down"));
+
+            await dataRefreshWatchdog.Invoking(w => w.RecoverStalledRefreshes()).Should().ThrowAsync<AggregateException>();
+
+            await serviceBusClient.Received(1).PublishToRequestTopic(Arg.Is<ValidatedDataRefreshRequest>(r => r.DataRefreshRecordId == 1));
```
SergeySalinAtDataart, line 171: I think this test should have also checked for the logger event like how the previous one checked for the absence of the event, because if you check for `.PublishToRequestTopic...(.DataRefreshRecordId == 2);` test will still pass

Atlas.MatchingAlgorithm.Test/Services/DataRefresh/DataRefreshWatchdogTests.cs
```diff
@@ -0,0 +1,242 @@
...
+            await serviceBusClient.Received(1).PublishToRequestTopic(Arg.Is<ValidatedDataRefreshRequest>(r => r.DataRefreshRecordId == 3));
+        }
+
+        [Test]
+        public async Task RecoverStalledRefreshes_WhenOneRecordFails_ThrowsOnlyOnceTheSweepIsComplete()
+        {
+            GivenStalledRecords(1, 2);
+            serviceBusClient.PublishToRequestTopic(Arg.Is<ValidatedDataRefreshRequest>(r => r.DataRefreshRecordId == 1))
```
SergeySalinAtDataart: What if several records fail? Wil it throw several times? I think we need more records, more of them should throw, but we need to assert that the method throws a single exception and it has several inner exceptions

Atlas.MatchingAlgorithm.Data.Persistent/Repositories/DataRefreshHistoryRepository.cs
```diff
@@ -164,6 +178,17 @@ public async Task<bool> ReleaseRefreshLease(int recordId, Guid owner)
             return rowsUpdated == 1;
         }
 
+        /// <inheritdoc />
+        public async Task<IReadOnlyCollection<int>> GetStalledRefreshRecordIds(DateTime graceCutoffUtc)
+        {
+            return await Context.DataRefreshRecords
+                .Where(r => r.RefreshEndUtc == null
+                            && (r.LeaseExpiresUtc == null || r.LeaseExpiresUtc < graceCutoffUtc)
+                            && (r.RefreshLastContinuedUtc ?? r.RefreshRequestedUtc) < graceCutoffUtc)
```
SergeySalinAtDataart, line 192: I've replaced this line with `&& r.RefreshLastContinuedUtc < graceCutoffUtc)`  and all the tests still pass 

Image

despite tests stating that they test for COALESCE. Curious. I think we need to add a new test to test the COALESCE

Atlas.MatchingAlgorithm.Test.Integration/IntegrationTests/DataRefresh/DataRefreshStalledRecordTests.cs
```diff
@@ -0,0 +1,166 @@
...
+
+            stalled.Should().Contain(recordId);
+        }
+
+        [Test]
+        public async Task GetStalledRefreshRecordIds_WhenNeverLeasedAndIdleBeyondTheGracePeriod_ReturnsRecord()
+        {
+            // The one stall no lease can reveal: a request message that was never delivered at all, leaving a record
+            // that has never been leased and never will be.
+            var recordId = await AnOpenRecordLastActive(TimeSpan.FromHours(3));
```
SergeySalinAtDataart, line 102: This test sets RefreshLastContinuedUtc so the comment actually tricks us. Is this like an abandoned run or something? The test is fine but the wording and premise need work

Atlas.MatchingAlgorithm.Data.Persistent/Repositories/DataRefreshHistoryRepository.cs
```diff
@@ -164,6 +178,17 @@ public async Task<bool> ReleaseRefreshLease(int recordId, Guid owner)
             return rowsUpdated == 1;
         }
 
+        /// <inheritdoc />
+        public async Task<IReadOnlyCollection<int>> GetStalledRefreshRecordIds(DateTime graceCutoffUtc)
+        {
+            return await Context.DataRefreshRecords
+                .Where(r => r.RefreshEndUtc == null
+                            && (r.LeaseExpiresUtc == null || r.LeaseExpiresUtc < graceCutoffUtc)
```
SergeySalinAtDataart: Here, we check whether LeaseExpiresUtc is null, but in TryClaimRefreshLease we never check if it is null. Probably worth a look - if somehow LeaseOwner is set but LeaseExpiresUtc is not, GetStalledRefreshRecordIds will pick it up, submit it for redelivery but it will never be claimed? (Maybe also worth a test, not sure)

terraform/core/variables.tf
```diff
@@ -390,6 +390,18 @@ variable "MATCHING_DATA_REFRESH_CRONTAB" {
   description = "A crontab determining when the matching data refresh will be auto-attempted. It will only run to completion if new HLA nomenclature is detected."
 }
 
+variable "MATCHING_DATA_REFRESH_WATCHDOG_CRON_SCHEDULE" {
+  type        = string
+  default     = "0 */15 * * * *"
+  description = "CRON schedule (six-field NCrontab) for the sweep that re-requests stalled data refreshes. Must be set: the Functions host resolves it at start-up, so the Data Refresh app will not start without it."
```
SergeySalinAtDataart: I might be wrong, but I think that the Data Refresh app will start up but will spam errors about this variable not set; at least, that's how it does it locally (not sure about actuel Azure environment though)

Atlas.MatchingAlgorithm/Services/DataRefresh/DataRefreshWatchdog.cs
```diff
@@ -0,0 +1,127 @@
...
+            logger.SendEvent(StallRecoveredEventName, LogLevel.Warn, new Dictionary<string, string>
+            {
+                { "DataRefreshRecordId", recordId.ToString() },
+                { "GraceCutoffUtc", graceCutoffUtc.ToString("u") }
+            });
+        }
+
+        private TimeSpan ValidatedGraceDuration()
+        {
+            var graceDuration = TimeSpan.FromMinutes(dataRefreshSettings.WatchdogGraceDurationMinutes);
```
SergeySalinAtDataart, line 112: I don't quite like this method, though it's probably fine for a function. If you have any ideas on how to improve settings' validation, I'm all ears though

SergeySalinAtDataart (changes requested):

da-ikovalenko (commented):
Automated code review findings (Claude Code).

Atlas.MatchingAlgorithm/Services/DataRefresh/DataRefreshWatchdog.cs
```diff
@@ -0,0 +1,127 @@
...
+            DataRefreshSettings dataRefreshSettings,
+            IMatchingAlgorithmImportLogger logger)
+        {
+            this.dataRefreshHistoryRepository = dataRefreshHistoryRepository;
+            this.serviceBusClient = serviceBusClient;
+            this.dataRefreshSettings = dataRefreshSettings;
+            this.logger = logger;
+        }
+
+        public async Task RecoverStalledRefreshes()
```
da-ikovalenko, line 60: **Correctness (plausible):** Re-publishing a stalled record doesn't mark it as "recently requested", so if the republished message isn't consumed before the next sweep, the same record is re-detected as stalled and re-published every sweep interval.

A record that fails processing repeatedly (e.g. a config/data bug that throws before `OrchestrateDataRefresh`'s `TryClaimRefreshLease`, or Service Bus/consumer backlog delays delivery past 15 minutes) never gets `LeaseExpiresUtc`/`RefreshLastContinuedUtc` updated merely from publishing, so `GetStalledRefreshRecordIds` flags it as stalled on every subsequent 15-minute sweep — duplicate `ValidatedDataRefreshRequest` messages pile up on the topic and a duplicate "Data refresh stall auto-recovered" event fires every cycle for what is really one ongoing (or repeatedly failing) situation.

README_Support.md
```diff
@@ -112,6 +121,22 @@ nothing left to replay. If so, publish one to the `data-refresh-requests` topic:
...
 
+If the record still holds a live lease, that request will be refused, and - **only if you are certain the
+original process is gone** - the lease has to be cleared first:
+
+```sql
+UPDATE [MatchingAlgorithmPersistent].[DataRefreshHistory]
+SET LeaseOwner = NULL, LeaseExpiresUtc = NULL
+WHERE Id = <record id>
+```
+
```
da-ikovalenko, line 134: **Docs (plausible):** The guidance to check for the "Data refresh stall auto-recovered" event "before assuming" the watchdog isn't running is weak — that event only fires when a stall is found and recovered, so its *absence* is also the normal healthy state and proves nothing about whether the timer function is actually executing. An on-call engineer could misread "no recovery event" as "watchdog not running" when it's simply the common case of nothing to do, and chase the wrong lead instead of checking the timer function's own invocation/exception telemetry.

Atlas.MatchingAlgorithm/Services/DataRefresh/DataRefreshWatchdog.cs
```diff
@@ -0,0 +1,127 @@
...
+
+            // Sent only once the publish has succeeded, so the event means "a request is on the topic", not "one was attempted".
+            logger.SendEvent(StallRecoveredEventName, LogLevel.Warn, new Dictionary<string, string>
+            {
+                { "DataRefreshRecordId", recordId.ToString() },
+                { "GraceCutoffUtc", graceCutoffUtc.ToString("u") }
+            });
+        }
+
+        private TimeSpan ValidatedGraceDuration()
```
da-ikovalenko, line 110: **Reuse (plausible):** `ValidatedGraceDuration` duplicates the shape of `DataRefreshOrchestrator.ValidatedLeaseTimings` (line 122 of that file) — both convert two int settings to `TimeSpan`s and validate one against `DataRefreshSettings.LeaseDurationMinutes` with a bespoke throw of `InvalidDataRefreshConfigurationException`, without sharing code. A future change to the lease-duration invariant has to be remembered and applied in both places.

Atlas.MatchingAlgorithm/Services/DataRefresh/DataRefreshWatchdog.cs
da-ikovalenko, line 60: **Reuse (plausible):** `RecoverStalledRefreshes` hand-rolls a "collect exceptions per item, keep going, throw `AggregateException` at the end" sweep pattern that already exists near-identically in `Atlas.Functions/Functions/ParallelMatchPredictionAggregatorFunctions.cs` (`FinaliseRuns`/`MarkRunsAsAbandoned`). This is now the third copy of the same shape in the codebase — a behavioural fix to it needs applying in three places.

Atlas.MatchingAlgorithm/Services/DataRefresh/DataRefreshWatchdog.cs
```diff
@@ -0,0 +1,127 @@
...
+            this.logger = logger;
+        }
+
+        public async Task RecoverStalledRefreshes()
+        {
+            var graceCutoffUtc = DateTime.UtcNow - ValidatedGraceDuration();
+
+            var stalledRecordIds = await dataRefreshHistoryRepository.GetStalledRefreshRecordIds(graceCutoffUtc);
+            if (stalledRecordIds.Count == 0)
+            {
```
da-ikovalenko, line 66: **Efficiency (plausible):** Stalled records are re-requested strictly sequentially via `await` in a `foreach`, even though each publish is independent and `PublishToRequestTopic` already has its own internal retry-with-cooldown (default 5 retries x 20s cooldown, up to ~100s per record). During a transient Service Bus outage with, say, 10 stalled records, sequential execution could take ~1000s (16+ minutes) — longer than the default 15-minute sweep interval — needlessly serializing independent I/O that `Task.WhenAll` could run concurrently.

Atlas.MatchingAlgorithm.Functions.DataRefresh/Functions/DataRefreshFunctions.cs
```diff
@@ -64,6 +67,18 @@ public async Task SubmitDataRefreshRequest([TimerTrigger("%DataRefresh:CronTab%"
             await dataRefreshRequester.RequestDataRefresh(request, false);
         }
 
+        /// <summary>
+        /// Re-requests every data refresh that has stalled, i.e. that is still open but that nothing is working on.
+/// Such a record blocks all subsequent refreshes until it is completed; the watchdog re-requests it automatically.
```
da-ikovalenko: **Style (confirmed):** This XML doc comment's second line ("Such a record blocks...") has zero leading whitespace, unlike every other `///` line in the file. Cosmetic only, but worth fixing before merge.

da-ikovalenko (changes requested):

Atlas.MatchingAlgorithm.Functions.DataRefresh/Functions/DataRefreshFunctions.cs
OTarasov-DataArt: Done, attribute dropped, parameter is now _.

Atlas.MatchingAlgorithm.Test/Services/DataRefresh/DataRefreshWatchdogTests.cs
OTarasov-DataArt: Agreed, removed. Renamed the survivor to RecoverStalledRefreshes_ReRequestsEveryStalledRecord since it's now the only test of that behaviour

Atlas.MatchingAlgorithm.Test/Services/DataRefresh/DataRefreshWatchdogTests.cs
OTarasov-DataArt: You're right, and I confirmed it, asserting == 2 passes too, since NSubstitute records the call even when it threw. Kept the Received() checks and added the recovery event for 1 and 3, plus its absence for 2. The event is only sent after the publish returns, so that's the part that proves they got through

Atlas.MatchingAlgorithm.Test/Services/DataRefresh/DataRefreshWatchdogTests.cs
OTarasov-DataArt: Done, three records, 1 and 3 throw and 2 succeeds, asserting one AggregateException with two inner exceptions. Verified by making the sweep break after the first failure, that now fails the test, which the single-failure version couldn't catch

Atlas.MatchingAlgorithm.Data.Persistent/Repositories/DataRefreshHistoryRepository.cs
OTarasov-DataArt: Yeah, I reproduced your edit and all 8 passed. My helper set RefreshLastContinuedUtc in every test expecting a record back, so the fall back was never exercised. Added WhenRequestedButNeverRunBeyondTheGracePeriod, which leaves it null, it fails with your edit applied.

Atlas.MatchingAlgorithm.Test.Integration/IntegrationTests/DataRefresh/DataRefreshStalledRecordTests.cs
OTarasov-DataArt: Yeah, that one was a bit misleading. RefreshLastContinuedUtc only gets stamped once a run has actually started, so what I'd written was an abandoned run, not an undelivered message. Renamed it to WhenARunWasAbandonedBeyondTheGracePeriod and added a proper never-run case alongside it

Atlas.MatchingAlgorithm.Data.Persistent/Repositories/DataRefreshHistoryRepository.cs
OTarasov-DataArt: Yeah, that seems like a possible outcome. Switched the clause to key off LeaseOwner so the two agree, and added a test for it

terraform/core/variables.tf
OTarasov-DataArt: You're right, I had this wrong. Ran func start without the setting and the host comes up fine, other six functions all bound, just RecoverStalledDataRefreshes showing as "in error". Only thing I'd add is that I left it running 4½ minutes and got 5 lines inside a ~120ms burst at startup and then absolutely nothing after. The host logs Function 'Functions.RecoverStalledDataRefreshes' failed indexing and will be disabled, so it's disabled outright rather than retried, which is why it goes quiet. Either way it fails quietly rather than loudly, which is probably worse, so I've reworded the XML comment and the Terraform description to say that

Atlas.MatchingAlgorithm/Services/DataRefresh/DataRefreshWatchdog.cs
OTarasov-DataArt: It's shaped that way to match ValidatedLeaseTimings() in the orchestrator. I think the better way would be .Validate(...).ValidateOnStart() on the options registration. RegisterAsOptions already builds on AddOptions<T>(), so one loud failure at boot instead of a check every tick. Nothing in the repo does that yet afaik and it'd mean moving the lease settings too, so I'm not sure if it's the right time and place for it

Atlas.MatchingAlgorithm.Functions.DataRefresh/Functions/DataRefreshFunctions.cs
OTarasov-DataArt: Already sorted, that was a GitHub suggestion I committed that lost its indentation, fixed in a later push

Atlas.MatchingAlgorithm/Services/DataRefresh/DataRefreshWatchdog.cs
OTarasov-DataArt: The timings look right. What I'm unsure about is whether we can get that many stalled records at once, RequestDataRefresh refuses to create one while any record is still open, so I think the sweep normally sees 0 or 1. Is there a path to 10 I'm missing? Happy to move to Task.WhenAll if so, though the failures list would need making thread-safe

Atlas.MatchingAlgorithm/Services/DataRefresh/DataRefreshWatchdog.cs
OTarasov-DataArt: Agreed, it is the third copy,  I mirrored MarkRunsAsAbandoned because the ticket asked for it, but that does leave the shape duplicated. Atlas.Functions already references Atlas.MatchingAlgorithm so placing a helper wouldn't be hard, it's more that it'd touch the two existing copies too, which felt beyond this ticket. Should I do it here?

Atlas.MatchingAlgorithm/Services/DataRefresh/DataRefreshWatchdog.cs
OTarasov-DataArt: They do look alike. The part I wasn't sure about is that they check different invariants, ValidatedLeaseTimings is renewal×2 ≤ lease, mine is grace > lease, so I wasn't sure a shared helper would save much beyond the exception type. Sergey flagged the same method on another thread and I suggested moving both to .Validate(...).ValidateOnStart() on the options registration instead. Does that seem a better direction than extracting something here?

README_Support.md
OTarasov-DataArt: You're right, I hadn't thought about it from the on-call side. Reworded it to point at the function's own invocation/exception telemetry, and to say explicitly that no event is the normal healthy state

Atlas.MatchingAlgorithm/Services/DataRefresh/DataRefreshWatchdog.cs
OTarasov-DataArt: You're right, and I think it's the same gap Copilot flagged twice. Publishing writes nothing to the record, so it stays eligible until something else claims it. The config case is the worst, ValidatedLeaseTimings() throws before TryClaimRefreshLease, so it'd never settle. The fix I can see is a column recording the last re-request so the predicate can back off, but that's a migration on DataRefreshHistory and outside what the ticket scoped. Would you rather I did that here?

Atlas.MatchingAlgorithm.Data.Persistent/Repositories/DataRefreshHistoryRepository.cs
SergeySalinAtDataart: If we can guarantee that if LeaseOwner is set then r.LeaseExpiresUtc is also set then it is fine, but so far this looks quite strange to me to check one property and then to compare another. Don't you agree? I felt that maybe a better solution was to add a check for LeaseExpiresUtc == null in TryClaimRefreshLease or something?

Atlas.MatchingAlgorithm/Services/DataRefresh/DataRefreshWatchdog.cs
SergeySalinAtDataart: Let's leave it like this. It is not time to make settings beautiful yet

SergeySalinAtDataart (approved):

Atlas.MatchingAlgorithm.Data.Persistent/Repositories/DataRefreshHistoryRepository.cs
OTarasov-DataArt: I think we can guarantee it, claim writes both columns in one ExecuteUpdateAsync, renew requires the owner, release nulls both. So it's only reachable via a manual DB edit.
The shape came from TryClaimRefreshLease already doing the same mix, minus the == owner case, I agree it reads oddly on its own. My only hesitation with your version is that it makes claiming more permissive, and I'm not sure which side is safer to fail on. Happy to go your way if you think it's clearer, though it would touch ATL-258's claim logic, which the ticket puts out of scope

da-ikovalenko (approved):

PR #1533, opened 2026-09-23
feature: ATL-272: implemented genotype imputation precompute service

SergeySalinAtDataart:
This change is [Reviewable](https://reviewable.io/reviews/Anthony-Nolan/Atlas/1533)


Atlas.MatchingAlgorithm.Data/Repositories/Precompute/SubjectGenotypeSetRepository.cs
```diff
@@ -0,0 +1,392 @@
...
+    /// </remarks>
+    Task<IReadOnlyDictionary<SubjectGenotypeSetKey, int>> GetOrCreateValueIds(IReadOnlyCollection<SubjectGenotypeSetValueToStore> values);
+
+    /// <summary>
+    /// Writes the per-donor mapping rows pointing donors at the values they resolve to.
+    /// </summary>
+    /// <remarks>
+    /// A plain insert. <c>IX_DonorSubjectGenotypeSets_DonorId_AllowedLociKey</c> is unique, so writing a donor that
+    /// already has a row for that combination throws - which is correct while the only caller is a refresh running
+    /// against freshly cleared tables.
```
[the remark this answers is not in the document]
SergeySalinAtDataart: Table cleanup will be implemented in [this ticket](https://anthonynolan.atlassian.net/browse/ATL-231)

Atlas.MatchingAlgorithm/DependencyInjection/ServiceConfiguration.cs
```diff
@@ -316,6 +318,14 @@ private static void RegisterDataRefreshServices(
             services.AddScoped<IDonorImporter, DonorImporter>();
         }
 
+        /// <summary>
+        /// Register the genotype set precompute service.
+        /// </summary>
+        public static void RegisterSubjectGenotypeSetPrecompute(this IServiceCollection services)
+        {
+            services.AddScoped<ISubjectGenotypeSetPrecomputeService, SubjectGenotypeSetPrecomputeService>();
```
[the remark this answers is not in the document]
SergeySalinAtDataart: Ticket explicitly says to ship this with no consumers. In later tickets, the actual consumers will wire up the service when needed

Atlas.MatchingAlgorithm/Services/DataRefresh/Precompute/SubjectGenotypeSetPrecomputeService.cs
```diff
@@ -0,0 +1,155 @@
...
+    /// <remarks>
+    /// One at a time: each call is a full imputation, whose own peak memory is what caps the donor batch size in the
+    /// first place. Running the batch's imputations concurrently is a change to make deliberately, with a measurement,
+    /// not a detail of this loop.
+    /// </remarks>
+    private async Task<IReadOnlyCollection<SubjectGenotypeSetValueToStore>> Compute(
+        IReadOnlyCollection<PrecomputeRequest> requests,
+        string matchingAlgorithmHlaNomenclatureVersion)
+    {
+        var values = new List<SubjectGenotypeSetValueToStore>(requests.Count);
```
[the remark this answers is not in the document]
SergeySalinAtDataart: Valid. Fixed

Atlas.MatchingAlgorithm/Services/DataRefresh/Precompute/SubjectGenotypeSetPrecomputeService.cs
```diff
@@ -0,0 +1,155 @@
...
+        internal SubjectGenotypeSetKey Key { get; }
+
+        internal PrecomputeRequest(PrecomputeSubject subject, AllowedLociKey allowedLociKey)
+        {
+            Subject = subject;
+            AllowedLociKey = allowedLociKey;
+            Key = new SubjectGenotypeSetKey(
+                SubjectGenotypeSetKeyGenerator.GenerateHlaTypingKey(subject.HlaTyping, allowedLociKey),
+                subject.FrequencySet.Id,
+                allowedLociKey);
```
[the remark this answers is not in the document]
SergeySalinAtDataart: Nomenclature Version was removed from the key deliberately, and precompute will be cleaned-up as part of [this ticket](https://anthonynolan.atlassian.net/browse/ATL-231)

Atlas.MatchPrediction/Services/Precompute/SubjectGenotypeSetPayload.cs
```diff
@@ -0,0 +1,565 @@
...
+            var genotypes = ReadGenotypes(reader, genotypeCount, pool, wide);
+            var sumOfLikelihoods = ReadDecimal(reader);
+
+            return new SubjectGenotypeSet(isUnrepresented, genotypes, sumOfLikelihoods);
+        }
+        catch (EndOfStreamException exception)
+        {
+            throw new InvalidDataException("Payload ended before the body was fully read; it is truncated.", exception);
+        }
+        catch (Exception exception) when (exception is FormatException or IOException)
```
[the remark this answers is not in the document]
SergeySalinAtDataart: No, you're wrong, it derives from `SystemException` and not `IOException`. I looked into the source code and confirmed that no `IOException` is mentioned and inheritance chain is InvalidDataException <- SystemException <- Exception.

da-ikovalenko (commented):
Reviewed against ATL-272, including its 2026-09-16 revision that takes the nomenclature version out of the key.

**Acceptance criteria:** all met. All four `AllowedLoci` combinations are covered. The payload is pool/diff/gzip encoded, and unrepresented rows are stored with NULL data. A round-trip test covers exact `decimal` fidelity. Only one value row is stored per key, within a batch and across runs, including under concurrency. The key is `(HlaTyping, HFSetId, AllowedLociKey)`, and the service ships dark.

**Checked and fine:** the payload format (pool, index width, null marker, overrides, decimals, envelope and version bytes), the guarded insert with its locking and retry, and lower-case hex under the case-insensitive collation. It's also good to see this reuse the existing `IGenotypeSetService` rather than duplicate the pipeline.

Inline comments below, most important first:
1. **Blocking:** a blank (`""`) and a missing (`null`) position get the same key, but imputation handles them differently.
2. **Before this is wired in:** nothing wipes the precompute tables on refresh yet (ATL-338).
3. Architectural: the new Matching -> MatchPrediction dependency.
4. One failed imputation fails the whole batch.
5. Nit: an unused `using`.

Atlas.MatchingAlgorithm/Services/DataRefresh/Precompute/SubjectGenotypeSetKeyGenerator.cs
```diff
@@ -0,0 +1,100 @@
...
+        foreach (var locus in LocusOrder)
+        {
+            if (!allowedLoci.Contains(locus))
+            {
+                continue;
+            }
+
+            foreach (var position in Positions)
+            {
+                builder.Append(hlaTyping.GetPosition(locus, position)).Append(PositionSeparator);
```
da-ikovalenko: **`null` and `""` share a key here, but imputation treats them differently.**

The remark above `Canonicalise` says "imputation treats them the same", and `GenerateHlaTypingKey_ForNullAndEmptyAtTheSamePosition_ReturnsTheSameKey` enforces that. It isn't true:

- `CompressedPhenotypeConverter.cs:84` only skips a position when `hla == null`. A null position counts as untyped and matches anything.
- A `""` position is sent to `ConvertHlaWithLoggingAndRetryOnFailure`. The lookup fails, logs "HLA Conversion Failed" and returns no groups, so that position matches no haplotypes and the donor usually comes out unrepresented.

Take donor 1 with `C = null` and donor 2 with `C = ""`, otherwise identical. They get the same `ABCDrb1` key. `DistinctBy(request => request.Key)` imputes only the first one reached, and the other donor is given its result. Depending on batch order, one of them ends up with the wrong genotype set, so the outcome isn't predictable either.

Suggestion: keep the shared key, and convert `""` to `null` in the typing before it's passed to `GetGenotypeSet` in `SubjectGenotypeSetPrecomputeService.Compute`. The key and the imputation then agree. Also worth adding a test that the two cases impute identically. (Live search today would send `""` to the lookup and fail. If the team decides that's the bug, fixing it upstream works too.)

Atlas.MatchingAlgorithm.Data/Repositories/Precompute/SubjectGenotypeSetRepository.cs
```diff
@@ -0,0 +1,397 @@
...
+namespace Atlas.MatchingAlgorithm.Data.Repositories.Precompute;
+
+public interface ISubjectGenotypeSetRepository
+{
+    /// <summary>
+    /// The ids of the keys that are already stored. Keys with no row are simply absent from the result.
+    /// </summary>
+    /// <remarks>
+    /// Purely an optimisation for the caller: it lets an expensive imputation be skipped for a typing that some
+    /// earlier batch, or some earlier refresh, has already paid for. It is NOT how duplicate rows are prevented -
```
da-ikovalenko: **The wipe-per-refresh step (ATL-338) doesn't exist yet, and this comment contradicts it.**

The key leaves out the nomenclature version on the understanding that `SubjectGenotypeSetValues` and `DonorSubjectGenotypeSets` are wiped on every data refresh. Nothing in production code clears them yet: the `DonorImportRepository` truncation only covers `Donors`, `DonorManagementLogs` and the HLA tables. Once this is wired in, a refresh to a new nomenclature version on the same dormant database would:
- find last refresh's rows through `GetExistingValueIds`, skip imputation, and give donors payloads converted under the old version;
- fail in `WriteDonorAssignments` on the unique `(DonorId, AllowedLociKey)` index because of the leftover assignment rows.

It's harmless while the service ships dark. But the wipe needs an owning ticket, and it has to land before ATL-232/233 wire this in. Could you also reword "or some earlier refresh" here? Under the agreed design, rows only survive within one refresh.

Atlas.MatchingAlgorithm/Atlas.MatchingAlgorithm.csproj
```diff
@@ -38,6 +38,10 @@
         <ProjectReference Include="..\Atlas.MatchingAlgorithm.Data\Atlas.MatchingAlgorithm.Data.csproj"/>
         <ProjectReference Include="..\Atlas.MatchingAlgorithm.Common\Atlas.MatchingAlgorithm.Common.csproj"/>
         <ProjectReference Include="..\Atlas.MatchingAlgorithm.Data.Persistent\Atlas.MatchingAlgorithm.Data.Persistent.csproj"/>
+        <!-- The data refresh precomputes each donor's imputed genotype set, which is match prediction's own pipeline
+             (IGenotypeSetService) and its payload encoder. Acyclic: Atlas.MatchPrediction references only Common,
+             Client.Models, the HLA Metadata Dictionary and its own Data project. -->
+        <ProjectReference Include="..\Atlas.MatchPrediction\Atlas.MatchPrediction.csproj"/>
```
da-ikovalenko: **Architectural: Matching now depends on MatchPrediction's business logic.**

ATL-272 implies this, since it asks for the MatchPrediction pipeline to be reused, so I'm not objecting. It is a new dependency between components, though, and worth an explicit acknowledgement from the team, or perhaps a short ADR note. Once this is wired in, the Matching Functions app will also need MatchPrediction's registrations: `IGenotypeSetService`, `GenotypeImputationSettings`, the HF-set services and caches, and the HMD. None of these are registered on the Matching side yet. That's fine while it ships dark, but it would be good to note in ATL-232/233.

Atlas.MatchingAlgorithm/Services/DataRefresh/Precompute/SubjectGenotypeSetPrecomputeService.cs
```diff
@@ -0,0 +1,202 @@
...
+                // result is shared by all of them. It reaches the logs of the conversion step, and nothing else.
+                new SubjectFrequencySet(request.Subject.FrequencySet, $"donor {request.Subject.DonorId}"));
+
+            // A copy of the locus set rather than the shared one behind AllowedLociKey.ToLoci(): MatchPredictionParameters
+            // exposes it as a mutable ISet. The allocation is nothing against the imputation that follows it.
+            var parameters = new MatchPredictionParameters(
+                request.AllowedLociKey.ToLoci().ToHashSet(),
+                matchingAlgorithmHlaNomenclatureVersion);
+
+            var genotypeSet = await genotypeSetService.GetGenotypeSet(subjectData, parameters);
```
da-ikovalenko: **One failed imputation fails the whole batch.**

If `GetGenotypeSet` throws (a malformed typing, or an HMD failure that outlasts the retries), the exception fails the whole `Precompute` call for all 2,000 donors. Chunks already stored are kept, which is good, but there's no way to fail a single donor. That's acceptable while it ships dark. It should line up with the ATL-233/ATL-314 stage-55 design (PermanentlyFailed status and graceful degradation) when this is wired in. Could you either note that on the ticket, or add a `TODO` here pointing at it?

Atlas.MatchingAlgorithm/DependencyInjection/ServiceConfiguration.cs
```diff
@@ -46,6 +47,7 @@
 using Atlas.MatchingAlgorithm.Settings;
 using Atlas.MatchingAlgorithm.Settings.Azure;
 using Atlas.MatchingAlgorithm.Settings.ServiceBus;
+using Atlas.MatchPrediction.Services.MatchProbability;
```
da-ikovalenko: Nit: this `using` looks unused. `RegisterSubjectGenotypeSetPrecompute` only references types from `Services.DataRefresh.Precompute`.

da-ikovalenko (changes requested):

Atlas.MatchingAlgorithm.Data/Repositories/Precompute/SubjectGenotypeSetRepository.cs
da-ikovalenko: Correction to the above: the wipe already has an owner. It is an acceptance criterion on **ATL-231** (2026-09-16 revision): `DonorImportRepository.RemoveAllDonorInformation` must also truncate `SubjectGenotypeSetValues`/`DonorSubjectGenotypeSets`, verified by a test. That criterion is still open, so no new ticket is needed. The only concern is ordering: ATL-231's wipe must land before ATL-232/233 wire this service in. The request to reword "or some earlier refresh" in this doc comment still stands.

Atlas.MatchingAlgorithm/DependencyInjection/ServiceConfiguration.cs
SergeySalinAtDataart: Cleaned up

Atlas.MatchingAlgorithm/Services/DataRefresh/Precompute/SubjectGenotypeSetPrecomputeService.cs
SergeySalinAtDataart: Added a comment

Atlas.MatchingAlgorithm/Atlas.MatchingAlgorithm.csproj
SergeySalinAtDataart: I'd very much like to extract the genotype imputation to its own project but none of the tickets currently want to do this refactoring so I decided that this decision is fine for the time being

Atlas.MatchingAlgorithm/Services/DataRefresh/Precompute/SubjectGenotypeSetKeyGenerator.cs
SergeySalinAtDataart: Suggestion implemented

Atlas.MatchingAlgorithm.Data/Repositories/Precompute/SubjectGenotypeSetRepository.cs
SergeySalinAtDataart: Nice

da-ikovalenko (approved):
Re-reviewed at 748f578. All five review comments are fixed or answered, and CI passes. Approving.

- The empty-name fix is correct, and the two new tests cover it. You were right that the HLA Metadata Dictionary throws for an empty name (`HlaConverter.cs:85`). My first comment was wrong about that detail.
- Your note that `WriteDonorAssignments` is not safe to repeat yet is a good find. I added it to ATL-233 as a requirement and an acceptance criterion. In the stage-55 design, the risk comes from a resumed stage 55, a Service Bus message delivered twice, or a failed batch sent again, not from the `HlaProcessor` overlap. You might reword the code comment to match, but that is not blocking.

OTarasov-DataArt (approved):
> @OTarasov-DataArt reviewed 17 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/Anthony-Nolan/Atlas/1533)* status: all files reviewed, 10 unresolved discussions (waiting on da-ikovalenko and daria-sorokina-da).

PR #1535, opened 2026-09-23
fix: ATL-395: invalidate HLA Metadata Dictionary cache after recreation

OTarasov-DataArt:
This change is [Reviewable](https://reviewable.io/reviews/Anthony-Nolan/Atlas/1535)


Atlas.HlaMetadataDictionary/ExternalInterface/DependencyInjection/HlaMetadataDictionaryCacheInvalidationConfiguration.cs
```diff
@@ -0,0 +1,189 @@
...
+                    $"'{ConnectionStringKey}' and '{UpdatedTopicKey}' must both be configured: without them this app " +
+                    "cannot subscribe to HLA Metadata Dictionary updates, and would serve stale HLA metadata for the " +
+                    "full cache lifetime after the dictionary is recreated.");
+            }
+
+            var subscription = BuildSubscriptionNameForThisInstance();
+            var autoDeleteOnIdle = TimeSpan.FromMinutes(
+                int.TryParse(configuration[AutoDeleteOnIdleMinutesKey], out var configuredMinutes)
+                    ? configuredMinutes
+                    : DefaultAutoDeleteOnIdleMinutes);
```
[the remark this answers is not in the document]
OTarasov-DataArt: Good catch, fixed. The [Range(5, ...)] annotation never ran on this path (the subscribing apps don't bind the settings type at all), so a value under 5 reached Service Bus and failed startup with an opaque error. Now validated at parse time with a message naming the setting and the limit, plus tests. The three other locations you list don't do a raw config parse, so I've left them.

Atlas.HlaMetadataDictionary/ExternalInterface/HlaMetadataCacheInvalidator.cs
```diff
@@ -0,0 +1,58 @@
...
+
+        public HlaMetadataCacheInvalidator(IPersistentCacheProvider cacheProvider, IAtlasLogger logger)
+        {
+            this.cacheProvider = cacheProvider;
+            this.logger = logger;
+        }
+
+        public void InvalidateCaches(string hlaNomenclatureVersion)
+        {
+            cacheProvider.RemoveWhere(key => HlaVersionedCacheKey.Matches(key, hlaNomenclatureVersion));
```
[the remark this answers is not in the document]
OTarasov-DataArt: I don't think it applies here. IAppCache.GetOrAddAsync (LazyCache) commits the cache entry synchronously, holding an unresolved AsyncLazy, and only then awaits the factory, so the entry is present and evictable for the whole fetch, RemoveWhere removes it, and resolving the lazy afterwards writes nothing back. The late-commit behaviour you describe is MemoryCache.GetOrCreateAsync, which we don't use. I verified all five GetOrAdd/GetOrAddAsync overloads in these paths behave the same way. Added InvalidationDuringCacheFillTests to pin it down, since it's a LazyCache property rather than ours and would regress silently if the library changed.

Atlas.HlaMetadataDictionary/ExternalInterface/DependencyInjection/HlaMetadataDictionaryCacheInvalidationConfiguration.cs
```diff
@@ -0,0 +1,220 @@
...
+                    "cannot subscribe to HLA Metadata Dictionary updates, and would serve stale HLA metadata for the " +
+                    "full cache lifetime after the dictionary is recreated.");
+            }
+
+            var subscription = BuildSubscriptionNameForThisInstance();
+            var autoDeleteOnIdle = ReadAutoDeleteOnIdle(configuration[AutoDeleteOnIdleMinutesKey]);
+
+            EnsureSubscriptionExists(connectionString, topic, subscription, autoDeleteOnIdle);
+
+            return builder.AddInMemoryCollection(new[]
```
SergeySalinAtDataart: Are you sure this configuration will work for Azure Functions? Can you try running it locally maybe just for the purpose of making sure that the function picks up the settings and starts

Atlas.HlaMetadataDictionary/ExternalInterface/DependencyInjection/HlaMetadataDictionaryCacheInvalidationConfiguration.cs
```diff
@@ -0,0 +1,220 @@
...
+        /// to itself, and only the instance itself knows which instance it is - so this cannot be declared in
+        /// Terraform alongside the topic.
+        ///
+        /// <para>
+        /// Subscriptions are created with an auto-delete-on-idle window, so one belonging to an instance that has
+        /// since been scaled away removes itself. A live instance holds an open receiver, which counts as activity
+        /// and keeps its own subscription alive.
+        /// </para>
+        /// </remarks>
+        public static IConfigurationBuilder AddPerInstanceHlaMetadataDictionaryCacheInvalidationSubscription(
```
SergeySalinAtDataart: I didn't see any calls to this subscription from Atlas.MatchPrediction.Worker and I think they also use HMD. Shall we add something for them as well or is it handled in some other way that I don't see here?

Atlas.HlaMetadataDictionary/ExternalInterface/DependencyInjection/HlaMetadataDictionaryCacheInvalidationConfiguration.cs
```diff
@@ -0,0 +1,220 @@
...
+                    // the Functions host captures.
+                    (exception, _, attempt, _) => Console.WriteLine(
+                        $"Could not ensure Service Bus subscription '{subscription}' on topic '{topic}'; " +
+                        $"attempt {attempt}/{MaxSubscriptionCreationRetries}; exception: {exception}"));
+
+            try
+            {
+                retryPolicy.Execute(() =>
+                {
+                    if (administrationClient.SubscriptionExistsAsync(topic, subscription).GetAwaiter().GetResult())
```
SergeySalinAtDataart: Maybe you can use an async overload instead so you'll have to write .GetAwaiter().GetResult() only once for that exact method instead of two inner methods? Just a thought

Atlas.Common/Caching/CacheProviders.cs
```diff
@@ -16,14 +38,48 @@ public interface ICacheProvider
...
         public IAppCache Cache { get; }
-        public PersistentCacheProvider(IAppCache cache) { Cache = cache; }
+
+        /// <summary>
+        /// The cache <see cref="Cache"/> is built over, or <c>null</c> when it was built elsewhere. Eviction by
+        /// predicate needs to enumerate keys, which only this type can do.
+        /// </summary>
+        private readonly MemoryCache underlyingCache;
+
+        public PersistentCacheProvider(IAppCache cache, MemoryCache underlyingCache = null)
```
SergeySalinAtDataart: I'm not sure where to put it, but when we evict the cache for HMD, we only evict the actual data that was cached and now we expect the consumers to go and fetch new versions of HMD. However, as I understand, we have some table clients for that and they `might` just point to the old versions of the tables so the cache is populated with already old data again. Can you please check if that's the case and whether it is worth looking into?

Atlas.MatchingAlgorithm.Functions.DataRefresh/Functions/HlaMetadataDictionaryCacheFunctions.cs
```diff
@@ -0,0 +1,44 @@
...
+    public class HlaMetadataDictionaryCacheFunctions
+    {
+        private readonly IHlaMetadataCacheInvalidator cacheInvalidator;
+
+        public HlaMetadataDictionaryCacheFunctions(IHlaMetadataCacheInvalidator cacheInvalidator)
+        {
+            this.cacheInvalidator = cacheInvalidator;
+        }
+
+        [Function(nameof(InvalidateHlaMetadataDictionaryCache))]
```
SergeySalinAtDataart: Claude flagged me a thing - if we start a data refresh and then do a manual HMD refresh (`RefreshHlaMetadataDictionaryToSpecificVersion`) we might have two workers competing for the same thing. Do you think we might want to protect it in scope of this ticket because it only costs us to check if there are any data refreshes underway?

Atlas.MatchingAlgorithm.Functions/Functions/HlaMetadataDictionaryCacheFunctions.cs
```diff
@@ -0,0 +1,44 @@
...
+        public HlaMetadataDictionaryCacheFunctions(IHlaMetadataCacheInvalidator cacheInvalidator)
+        {
+            this.cacheInvalidator = cacheInvalidator;
+        }
+
+        [Function(nameof(InvalidateHlaMetadataDictionaryCache))]
+        public void InvalidateHlaMetadataDictionaryCache(
+            [ServiceBusTrigger(
+                "%HlaMetadataDictionaryNotifications:UpdatedTopic%",
+                "%HlaMetadataDictionaryNotifications:UpdatedSubscription%",
```
da-ikovalenko: **Blocking:** I think this binding expression will not resolve.

These are isolated-worker apps. In this model the Functions **host** process (not the worker) reads the trigger metadata and resolves `%...%` names from its own app settings. `AddPerInstanceHlaMetadataDictionaryCacheInvalidationSubscription` puts `UpdatedSubscription` only into the worker's configuration (`AddInMemoryCollection`). Terraform and `local.settings.json` do not set it. So the host cannot see the value. The function will fail to load, or it will never listen. Each instance creates a subscription that nobody reads, and it deletes itself after 60 minutes idle. The scale controller also reads trigger settings from app settings, so it has the same problem.

The same code is in the DonorManagement and DataRefresh copies of this file.

One option: do not use `[ServiceBusTrigger]` here. Start a receiver from the worker itself, in a hosted service, using the `Atlas.Common/ServiceBus` receiver abstractions. Then the per-instance subscription name only needs to exist inside the worker.

Atlas.HlaMetadataDictionary/ExternalInterface/DependencyInjection/HlaMetadataDictionaryCacheInvalidationConfiguration.cs
da-ikovalenko: This value only exists in the worker's configuration. The Functions host cannot see it, so the `%HlaMetadataDictionaryNotifications:UpdatedSubscription%` trigger binding will not resolve. See my comment on `HlaMetadataDictionaryCacheFunctions.cs`. The tests for this class only check configuration inside the same process, so they cannot catch this.

Atlas.HlaMetadataDictionary/ExternalInterface/DependencyInjection/HlaMetadataDictionaryCacheInvalidationConfiguration.cs
```diff
@@ -0,0 +1,220 @@
...
+
+        /// <remarks>
+        /// <c>WEBSITE_INSTANCE_ID</c> is a 64 character hex id and Service Bus caps subscription names at 50, so it is
+        /// hashed rather than truncated - the ids of two instances can share a long common prefix. Outside Azure
+        /// (local development) there is no such id, so the machine and process stand in, which also lets two locally
+        /// running apps hold separate subscriptions.
+        /// </remarks>
+        internal static string BuildSubscriptionNameForThisInstance()
+        {
+            var instanceIdentity = Environment.GetEnvironmentVariable("WEBSITE_INSTANCE_ID");
```
da-ikovalenko: Minor, not blocking: the name is a hash of `WEBSITE_INSTANCE_ID` only. If two different apps ever get the same instance id on the same plan VM, they would share one subscription and compete for messages. Adding `WEBSITE_SITE_NAME` to the hashed value costs nothing and removes this risk.

Atlas.HlaMetadataDictionary/ExternalInterface/HlaMetadataCacheInvalidator.cs
da-ikovalenko: **Medium:** stale data can be cached again right after this eviction.

`TableClientRepositoryBase` keeps a `TableClient` per table prefix in its `tableClients` field (`CloudTableRepositoryBase.cs:39`). Only `RecreateDataTable` clears it, and that runs in the **publishing** process, not here. Recreation writes a new physical table and does not delete the old one.

Scenario: a search or donor batch got its `IHlaMetadataDictionary` from the Factory cache before the notification arrived. Its repositories still point at the old table. After this eviction it misses the cache and calls `FetchAllRowsInTable` on the old table. The old rows are stored again under the same versioned key, for another 24h.

New scopes get fresh repositories, so the window is small, but it is real. `InvalidationDuringCacheFillTests` covers fills that are already running at eviction time. It does not cover fills that start after eviction from objects that still point at the old table.

Atlas.HlaMetadataDictionary/ExternalInterface/HlaMetadataDictionary.cs
```diff
@@ -161,6 +165,12 @@ public async Task<string> RecreateHlaMetadataDictionary(CreationBehaviour recrea
                 logger.SendTrace($"HLA-METADATA-DICTIONARY REFRESH: Recreating HLA Metadata dictionary for desired HLA Nomenclature version.");
                 await recreateMetadataService.RefreshAllHlaMetadata(version);
                 logger.SendTrace($"HLA-METADATA-DICTIONARY REFRESH: HLA Metadata dictionary recreated at HLA Nomenclature version: {version}");
+
+                // Only on the branch that actually rewrote storage, and only once storage has been rewritten: a
+                // consumer that clears its cache before the new data is in place would simply re-cache the old data.
+                // Both recreation routes pass through here, which is what keeps the forced same-version refresh and a
+                // normal version-changing data refresh behaving identically.
+                await updateNotifier.NotifyOfUpdate(version);
```
da-ikovalenko: **Low/medium:** this call is not guarded. If Service Bus is down, the send fails after about 5 × 20s of retries, and the exception leaves `RecreateHlaMetadataDictionary` **after** storage has already been rewritten.

In `DataRefreshRunner.RefreshHlaMetadataDictionary` this happens before `UpdateExecutionDetails` and `MarkStageAsComplete`. So the refresh fails at its first stage, and a retry repeats the whole long recreation. The manual `RefreshHlaMetadataDictionary(ToSpecificVersion)` endpoints also report failure, but the recreation worked.

Suggestion: catch and log the error here (a failed notification only means consumers stay stale until the TTL ends), or record the stage as complete before sending the notification.

da-ikovalenko (commented):
Thanks for this. The design matches ATL-395 well: version-scoped eviction, a no-op notifier by default, and Manage rights limited to one topic are all good choices.

There is one blocking problem (see the comment on `HlaMetadataDictionaryCacheFunctions.cs`). I think the trigger will not run in Azure, so the stale-cache bug would stay. Could you please show one real end-to-end run? Recreate the dictionary at the same version, then show the invalidation trace in App Insights for every running instance.

**Check against the ticket:** the ATL-395 acceptance criteria say "all already-running consuming Function Apps". This PR only covers the three Matching Algorithm apps. The README says that `Atlas.Functions`, `MatchPrediction.Functions` and `RepeatSearch.Functions` stay stale for up to 24h, on purpose. That is a fair scope choice, but please write it on the ticket, or link a follow-up ticket.

da-ikovalenko (changes requested):

da-ikovalenko:
Thanks for testing this locally and reporting the trigger problem. You are right that the `ServiceBusTrigger` cannot work here: the host resolves `%...%` binding expressions from app settings, and the per-instance subscription name only exists later, in the worker.

I suggest we don't move to a background `ServiceBusProcessor` either. The real need is that a message reaches **every instance** of an app, and a Service Bus subscription gives each message to only **one** receiver. Polling a shared marker fits this better.

## Proposal: poll a "recreation stamp" instead of using Service Bus

**The idea.** Each time the dictionary is recreated, it records a new stamp for that nomenclature version in storage. Each running worker instance checks the stamps every minute or so. If the stamp for a version has changed since the last check, the instance evicts that version from its cache.

Every instance reads the same stamp, so every instance finds out. Nothing is created per instance, so the problems you listed go away.

### How it works

1. **Writing the stamp.** In `RecreateHlaMetadataDictionary`, at the point where the PR now calls `updateNotifier.NotifyOfUpdate(version)`, write or update one row in a small new HMD table (for example `HlaMetadataDictionaryRecreations`): key = version, value = a new GUID plus `RecreatedAtUtc`.
   - It uses the same storage account and the same connection the HMD already uses.
   - No new credentials, and no opt-in registration. Every app that can recreate the dictionary writes the stamp automatically, including `Atlas.MatchingAlgorithm.Api`, which the PR currently leaves out.
2. **Local eviction.** Right after writing the stamp, the recreating process calls `IHlaMetadataCacheInvalidator.InvalidateCaches(version)` on itself. That instance is correct at once and does not wait for a poll.
3. **Polling.** A new `HlaMetadataDictionaryRecreationWatcher : BackgroundService` runs in each consuming app. It is registered with `AddHostedService`. Hosted services already run in these isolated workers (see `TelemetryFlushService`, registered in `LoggerRegistration.cs`). The watcher:
   - reads all stamp rows every N seconds (default 60, configurable). That is one small table query per instance per minute, so the cost is very small.
   - keeps the stamps it last saw in memory, and calls `InvalidateCaches(version)` for any stamp that has changed.
   - on its first read, only records the stamps as the starting point. A new process has an empty cache, so there is nothing to evict.
   - catches and logs errors (warning level) and tries again on the next tick. **Startup never depends on the stamp table**, and if storage is down the cache simply stays as it is.
4. **Scale-in and scale-out need nothing.** A new instance starts with an empty cache. A removed instance leaves nothing behind.

### Comparison with the background-processor fix

| Concern | Per-instance Service Bus processor | Stamp polling |
|---|---|---|
| Every instance gets told | Yes, but only through per-instance subscriptions | Yes, by design |
| Rights needed | Manage on the topic | Existing HMD table access |
| Entities created at runtime | Yes, plus auto-delete for removed instances | None |
| Startup depends on Service Bus | Yes | No |
| Silent failure | Possible (as you found) | Failures are logged each tick, and a new instance's first poll proves the table is readable |
| Infrastructure | Topic, auth rule, audit subscription, app settings per app | One table, created on first write |
| Delay before caches are correct | Seconds | Up to the poll interval (~1 min) |

The one cost is up to one minute of delay. Today the cache can be stale for up to 24 hours, and recreation is rare and started by operators, so one minute is fine. The runbook for `RefreshHlaMetadataDictionaryToSpecificVersion` should say "wait about one minute before testing again".

### What stays from this PR

Most of the hard work stays as it is:
- `HlaVersionedCacheKey`, `IPersistentCacheProvider.RemoveWhere`, and the change in `CacheRegisteringExtensions` that keeps the underlying `MemoryCache`
- `HlaMetadataCacheInvalidator` and the version-scoped eviction, including the reasoning about keeping warm caches during a data refresh
- the notify point in `RecreateHlaMetadataDictionary`, which now writes the stamp instead of publishing
- the cache-key tests: `HlaVersionedCacheKeyTests`, `InvalidationDuringCacheFillTests`, `ScoringCacheInvalidationTests`

### What gets removed

- the `ServiceBusTrigger` functions in all three apps
- `HlaMetadataDictionaryCacheInvalidationConfiguration` and the `ConfigureAppConfiguration` calls
- `HlaMetadataDictionaryNotificationSettings`, the Service Bus notifier and the non-publishing notifier
- the `Azure.Messaging.ServiceBus` and `Microsoft.Extensions.Configuration` package references in the HMD project
- the topic, the Manage auth rule and the audit subscription in `servicebus.tf`, plus the new `servicebus_topics` variable and the app settings
- the README text about Manage rights and topic setup (replace it with a short paragraph about the stamp)

The PR gets smaller overall. It adds one table repository, one hosted service and one poll-interval setting.

### Other consumers become easy to add

Today only the three Matching apps are covered. With polling, covering `Atlas.Functions`, `MatchPrediction.Functions` or `RepeatSearch.Functions` is one `AddHostedService` line. I suggest a registration helper such as `RegisterHlaMetadataDictionaryCacheInvalidation()`. We can add them in this PR or in a follow-up ticket.

### Points to keep in the new version

- **Write the stamp only after storage is fully rewritten.** The PR already does this. A consumer that evicts too early would just cache the old data again.
- **Do not fail the recreation if the stamp write fails.** The dictionary data is already written at that point. Log an error, and the operator can run it again. Otherwise a data refresh could fail after its first stage has already succeeded.
- **Watcher tests:** the first read only records the starting point; a changed stamp evicts only that version; an unchanged stamp does nothing; a read error is logged and the loop keeps running.

### Alternative considered, not recommended

**Azure App Configuration sentinel key.** The Matching Functions app already uses App Configuration, and a sentinel key with refresh is the standard way to tell all instances about a change. But the recreating apps would need write rights on App Configuration, and in isolated Functions the refresh runs through invocation middleware, so the check would run on the request path and not in the background. The HMD table approach keeps everything inside the HMD.

About asking earlier: the per-instance idea was a reasonable answer to a real problem (one subscription per app reaches only one instance), and the trigger-indexing failure is a platform limit that is easy to miss. Please add one line to the README that `ServiceBusTrigger` binding expressions are resolved by the host from app settings, not from worker configuration.

